PR #7, opened 2025-04-02
EU token support

petyosi:
@davidhewitt - I assume you did the same for the Rust SDK, so you can help me with the logic

### DIFF
--- a/packages/logfire/src/logfireConfig.ts
+++ b/packages/logfire/src/logfireConfig.ts
@@ -43,6 +43,11 @@ export interface SrubbingOptions {
   extraPatterns?: string[]
 }
 
+export interface RegionData {
+  baseUrl: string
+  gcpRegion: string
+}
+
 export interface LogfireConfigOptions {
   /**
    * Additional span processors to be added to the OpenTelemetry SDK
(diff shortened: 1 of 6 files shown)


davidhewitt (approved):

packages/logfire/src/logfireConfig.ts
```diff
@@ -161,8 +165,8 @@ export function configure(config: LogfireConfigOptions = {}) {
     logfireApi.configureLogfireApi({ otelScope })
   }
 
-  const baseUrl = resolveBaseUrl(cnf.advanced?.baseUrl)
   const token = cnf.token ?? env.LOGFIRE_TOKEN
+  const baseUrl = resolveBaseUrl(cnf.advanced?.baseUrl, token)
```
Viicos, line 169: Maybe instead what can be done:
- first check for `resolveSendToLogfire(cnf.sendToLogfire, token)`. If `false`, set `baseUrl` to an empty string (as it shouldn't be used).
- If `true`, assert that `token` is not undefined, and fail eagerly.
- Only then call `resolveBaseUrl()`

This way you can type the `token` parameter as `string` in `resolveBaseUrl()` and `getBaseUrlFromToken()`

PR #104, opened 2026-04-17
Add @pydantic/evals package - TypeScript port of pydantic-evals

Kludex:
## Summary

- Ports [pydantic-evals](https://ai.pydantic.dev/evals/) to TypeScript as a new `@pydantic/evals` package in this monorepo.
- Port covers the full surface: `Case`/`Dataset`, built-in evaluators (`Equals`, `EqualsExpected`, `Contains`, `IsInstance`, `MaxDuration`, `HasMatchingSpan`, `LLMJudge`), report-level evaluators (`ConfusionMatrixEvaluator`, `PrecisionRecallEvaluator`, `ROCAUCEvaluator`, `KolmogorovSmirnovEvaluator`), OTel span-tree capture via `getSpanTreeProcessor()`, YAML/JSON serialization, lifecycle hooks, online-evaluation decorator with sampling/sinks.
- Emits the same hierarchy of OTel spans as the Python package, so Logfire traces look the same.

## Reproducing the Python example in TypeScript

The TS snippet below is the exact counterpart of the Python script we integration-tested against (`~/dev/pydantic/platform/dataset_load.local.py`). Run it with a valid `LOGFIRE_TOKEN` + `LOGFIRE_BASE_URL` and you'll get the same 31-span trace (1 `experiment` + 1 `evaluate {name}` + 4 `case: {case_name}` + 4 `execute {task}` + 4 `task: classify` + 17 `evaluator: {evaluator_name}`) with identical attributes on each span kind.

### Python (for reference)

```python
import logfire
from pydantic_evals import Case, Dataset
from pydantic_evals.evaluators import Contains, EqualsExpected, LLMJudge

logfire.configure(base_url='http://localhost:3000', token='logfire-write-token')

async def classify(inputs: dict) -> str:
    with logfire.span('task: classify', text=inputs['text']):
        t = inputs['text'].lower()
        if 'error' in t or 'fail' in t: return 'NEGATIVE'
        if 'great' in t or 'love' in t: return 'POSITIVE'
        return 'NEUTRAL'

dataset = Dataset(
    name='sentiment-classifier',
    cases=[
        Case(name='positive-1', inputs={'text': 'I love this!'}, expected_output='POSITIVE'),
        Case(name='negative-1', inputs={'text': 'This is an error'}, expected_output='NEGATIVE'),
        Case(name='neutral-1',  inputs={'text': 'just fine'},        expected_output='NEUTRAL'),
        Case(
            name='great-with-contains',
            inputs={'text': 'it is great'},
            expected_output='POSITIVE',
            evaluators=(Contains(value='POSITIVE'),),
        ),
    ],
    evaluators=[EqualsExpected(), LLMJudge(rubric='Output is non-empty.')],
)

await dataset.evaluate(classify, max_concurrency=4)
```

### TypeScript equivalent (this PR)

```ts
import * as logfire from '@pydantic/logfire-node'
import {
  Case,
  Contains,
  Dataset,
  EqualsExpected,
  getSpanTreeProcessor,
  LLMJudge,
  setDefaultJudgeFn,
} from '@pydantic/evals'

// Plug the pydantic-evals span-tree capture into logfire's TracerProvider so
// `HasMatchingSpan` / ctx.spanTree queries work inside evaluators.
logfire.configure({
  serviceName: 'sentiment-classifier',
  additionalSpanProcessors: [getSpanTreeProcessor() as never],
})

// Bring-your-own judge: pydantic-evals doesn't bundle a model client.
setDefaultJudgeFn(async ({ output }) => ({
  pass_: typeof output === 'string' && output.length > 0,
  reason: 'output is non-empty',
  score: 0.9,
}))

async function classify(inputs: { text: string }): Promise<string> {
  return await logfire.span('task: classify', { text: inputs.text }, {}, async () => {
    const t = inputs.text.toLowerCase()
    if (t.includes('error') || t.includes('fail')) return 'NEGATIVE'
    if (t.includes('great') || t.includes('love')) return 'POSITIVE'
    return 'NEUTRAL'
  })
}

const dataset = new Dataset<{ text: string }, string>({
  name: 'sentiment-classifier',
  cases: [
    new Case({ name: 'positive-1', inputs: { text: 'I love this!' }, expectedOutput: 'POSITIVE' }),
    new Case({ name: 'negative-1', inputs: { text: 'This is an error' }, expectedOutput: 'NEGATIVE' }),
    new Case({ name: 'neutral-1',  inputs: { text: 'just fine' },         expectedOutput: 'NEUTRAL' }),
    new Case({
      name: 'great-with-contains',
      inputs: { text: 'it is great' },
      expectedOutput: 'POSITIVE',
      evaluators: [new Contains({ value: 'POSITIVE' })],
    }),
  ],
  evaluators: [new EqualsExpected(), new LLMJudge({ rubric: 'Output is non-empty.' })],
})

const report = await dataset.evaluate(classify, { maxConcurrency: 4 })
console.log(report.render({ includeInput: true, includeOutput: true }))
```

### Span parity (verified against local Logfire)

Running the Python and TypeScript versions side by side against the same local backend yields identical trace shape:

| | TS | PY |
|---|---|---|
| Total spans | **31** | **31** |
| `experiment: sentiment-classifier` | 1 | 1 |
| `evaluate {name}` | 1 | 1 |
| `case: {case_name}` | 4 | 4 |
| `execute {task}` | 4 | 4 |
| `task: classify` | 4 | 4 |
| `evaluator: {evaluator_name}` | 17 | 17 |
| `case` span attr keys | 12 (identical set) | 12 |
| `evaluate` span attr keys | 6 (identical set + `assertion_pass_rate` post-run) | 6 |

## JS-specific design choices

- `AsyncLocalStorage` replaces Python's `ContextVar` for the current task run.
- `getSpanTreeProcessor()` for OTel v2 since `addSpanProcessor` was removed (legacy v1 path preserved behind pragmas).
- Online decorator works on async functions only — sync background dispatch doesn't fit the JS concurrency model cleanly.
- `LLMJudge` takes a `judge` callback (or `setDefaultJudgeFn`) instead of bundling pydantic-ai.
- Rich-style terminal rendering replaced with plain-text tables.

## Test plan

- [x] 187 unit tests covering all public API surfaces (`pnpm test`)
- [x] 95% line coverage, 89% branch coverage, 98% function coverage (with strict thresholds in `vitest.config.ts`)
- [x] Full `pnpm check` (build + typecheck + lint + format + test) passes across all 5 packages
- [x] Integration-tested end-to-end against local Logfire backend - 31-span trace matches Python's output exactly

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## AI Disclaimer

This PR was developed with the assistance of either Claude or Codex.


petyosi:
Review note from looking at this PR as a possible direction for a JS evals package, especially with Logfire connectivity baked in.

I evaluated the final closed PR head (`95f4d39145cb9c4de86de5687c23c11abb57a75f`), not the intermediate commit that briefly included generated `dist/` and coverage output.

## Current Logfire / OTel integration in this PR

The current implementation is mostly generic OpenTelemetry with some Logfire-shaped attributes. It is not yet a first-class Logfire JS integration.

What exists:

- `packages/pydantic-evals/src/tracing.ts` dynamically imports `@opentelemetry/api`, gets a tracer named `pydantic-evals`, and creates spans for eval execution. These spans include `logfire.msg`, `logfire.msg_template`, and `logfire.span_type`.
- `Dataset.evaluate()` emits an `evaluate {name}` span with attributes like `dataset_name`, `task_name`, `n_cases`, `gen_ai.operation.name = "experiment"`, optional `metadata`, and `logfire.experiment.repeat`.
- Case spans emit attributes for inputs, expected output, output, metrics, attributes, assertions, scores, labels, and `logfire.experiment.source_case_name` when `repeat > 1`.
- `getSpanTreeProcessor()` can be passed manually to `@pydantic/logfire-node` via `logfire.configure({ additionalSpanProcessors: [getSpanTreeProcessor()] })`. That lets `HasMatchingSpan` inspect spans emitted during a task.
- The implementation extracts aggregate task metrics from captured GenAI spans when attributes such as `gen_ai.request.model`, `gen_ai.operation.name = "chat"`, `operation.cost`, and `gen_ai.usage.*` are present.
- Online evaluation has a `spanReference` field in the sink interface, but the implementation always passes `null`.

What does not exist yet:

- `@pydantic/evals` does not depend on or import `logfire` or `@pydantic/logfire-node`; it only has optional OTel peers.
- There is no first-class `logfire.configure()` integration, helper, preset, or default setup path.
- There is no default online-eval emission to Logfire or OTel events. If no sink is configured, online evaluators effectively have nowhere to send results.
- `traceId` and `spanId` fields on reports/cases are always `null`, despite spans being created.
- Report evaluators run after the main `evaluate {name}` span closes, so report analyses are not attached to the experiment span.
- The Python implementation sets experiment-level attributes such as `logfire.experiment.metadata`, `logfire.experiment.analyses`, and report evaluator failures; this PR only sets `assertion_pass_rate`.
- The tracing helper hand-rolls Logfire-style formatting and attribute normalization instead of using the existing `logfire` package formatting/scrubbing behavior.
- The package uses Node-only primitives (`node:async_hooks`, `setImmediate`) but the package boundary/docs do not clearly present this as a Node-only Logfire integration.

## Broader gaps before treating this as product direction

- Docs are effectively absent: the package README is five lines. A public package would need install/setup, Logfire setup, quick start, evaluator docs, online evaluation docs, serialization docs, examples, and known parity gaps.
- Python parity is partial. The PR covers the broad visible surface (`Case`, `Dataset`, built-in evaluators, report evaluators, span tree capture, lifecycle hooks, YAML/JSON text serialization, online eval), but misses or diverges on important behavior:
  - no `fromFile` / `toFile`;
  - no JSON schema generation for dataset files;
  - no retry support for task/evaluator execution;
  - `progress` exists in the TS API but is unused;
  - `LLMJudge` is bring-your-own callback only, not model-integrated;
  - no rich report renderer or equivalent configurable renderer;
  - no runtime validation comparable to Python's Pydantic models.
- TypeScript idiom is mixed:
  - constructor object parameters and generics are reasonable;
  - broad lint rule disabling, heavy casts, runtime reliance on `constructor.name` + `keepNames`, and weak deserialization validation are not great foundations;
  - the online decorator is typed as preserving the original function type, but sampled sync functions return a `Promise` at runtime.
- Monorepo fit is mechanical but incomplete:
  - `packages/*` picks it up;
  - it uses shared Vite/tsconfig shape;
  - but dependency versions are hard-coded instead of using the pnpm catalog;
  - no changeset;
  - no examples;
  - no docs-site integration;
  - lint config carves out many rules compared with neighboring packages.

## Suggested direction if Logfire connectivity is the goal

I would treat this PR as a prototype of the eval API shape, not as the architecture for Logfire-connected JS evals.

The next version should probably start from a Logfire-first contract:

- one clear setup path, e.g. `configureEvalsForLogfire()` or automatic integration when `@pydantic/logfire-node` is configured;
- default dataset/evaluator/online-eval telemetry that lands in Logfire without the user writing a custom sink;
- populated `traceId`/`spanId` on reports and cases;
- experiment metadata, analyses, and report evaluator failures attached to the experiment span;
- online eval result/failure events using the GenAI evaluation event model;
- span-tree capture installed by default when Logfire Node is used, with an escape hatch for custom OTel providers;
- reuse of the existing Logfire JS formatting/scrubbing/export behavior rather than parallel local implementations.

## Exact prompts used for this review

```text
Marcelo opened a PR for adding a JS port of pydantic evals, can you find it?
```

```text
he closed it subsequently
```

```text
Evaluate it: 
- Does it include the necessary docs? 
- Does it do feature parity with the python version?
- Does it follow idiomatic TypeScript? 
- Does it fit the current monorepo? 

Give me several comparisons of how a certain eval scenario would look like in Python and TypeScript
```

```text
Something which we want to take as a direction (a bit different than what's going on in Pydantic Evals) is to bake the Logfire connectivity in there. What are the current logfire js integrations in this PR?
```

PR #172, opened 2026-07-16
Expose OpenTelemetry SpanKind in manual span APIs

### LINKED ISSUE #171: Expose OpenTelemetry SpanKind in manual span APIs
## Problem

The manual Logfire JavaScript APIs do not expose OpenTelemetry's `SpanKind`.

The underlying OTel tracer accepts `SpanOptions.kind`, but `logfire.span()`, `logfire.startSpan()`, and `logfire.startPendingSpan()` currently construct their own span options containing only attributes. This means manually instrumented remote operations are always exported with OTel's default `INTERNAL` kind.

`SpanKind` must be provided when a span is created and cannot be corrected afterward. Calling the underlying tracer directly is possible, but requires consumers to duplicate Logfire's message formatting, attribute serialization, scrubbing, level, pending-span, and exception behavior.

## Motivating use case

The Logfire browser frontend manually instruments one-shot WebSocket requests to the backend. Each attempt is an outbound request/response operation and should be a `CLIENT` span, while connection milestones are span events.

The same gap applies to other manually instrumented operations:

- HTTP, RPC, and WebSocket clients
- server handlers not covered by auto-instrumentation
- queue producers and consumers
- background or deferred work

Accurate span kinds improve service topology, dependency analysis, and interoperability with OTel-aware backends.

## Proposed API

Expose an optional kind on the manual span APIs, following OTel's naming:

```ts
import { SpanKind } from '@opentelemetry/api'
import * as logfire from 'logfire'

await logfire.span('load live view records', {
  kind: SpanKind.CLIENT,
  attributes: {
    'websocket.endpoint': 'historical_query',
  },
  callback: async (span) => {
    // ...
  },
})
```

One possible implementation is:

- Add `kind?: SpanKind` to the options used by `span()`, `startSpan()`, and `startPendingSpan()`.
- Forward it to `Tracer.startSpan()` / `Tracer.startActiveSpan()`.
- Preserve the kind on pending-span placeholders.
- Consider exposing the same option through `instrument()`.
- Keep the default unchanged: omitting `kind` continues to produce `INTERNAL` spans.

The exact type organization may need care because `LogOptions` is also used by zero-duration Logfire logs; a dedicated span-options type would also solve the problem.

## Acceptance criteria

- Manual span APIs can create `CLIENT`, `SERVER`, `PRODUCER`, and `CONSUMER` spans.
- Existing calls remain source- and behavior-compatible.
- Pending and final representations use the same kind.
- Tests using an in-memory exporter verify the exported OTel kind.
- Documentation includes a manual remote-operation example.

ayaangazali:
This adds the optional `kind` from #171 to the manual span APIs. `span()`, `startSpan()` and `startPendingSpan()` accept `kind?: SpanKind` through a dedicated `SpanOptions` type that extends `LogOptions`, so the log helpers keep their existing options and can't pick up a kind. `instrument()` takes the same option via `InstrumentOptions`. The kind is forwarded to `Tracer.startSpan()` / `startActiveSpan()`, the manual pending placeholder in `startPendingSpanWithSettings` reuses it, and `PendingSpanProcessor` already copies `span.kind` onto automatic placeholders, so pending and final spans stay consistent on both paths. Omitting `kind` still produces `INTERNAL` spans.

Fixes #171

Tests: new `spanKind.integration.test.ts` with an in-memory exporter covering CLIENT/SERVER/PRODUCER/CONSUMER across both `span()` call shapes, `startSpan()`, `startPendingSpan()` (placeholder and real span get the same kind), `instrument()`, and the INTERNAL default. Docs example added to `docs/reference/api.md`, changeset included (minor for `logfire`). Ran `pnpm run check` from the root (build, vp check, typecheck, all package tests, release tooling), all green on Node 24.

quick note: I built this with Claude Code's help and reviewed the final diff myself. apologies if I missed something obvious, freshman in college trying my best to contribute, would genuinely love feedback :)

### DIFF
--- a/.changeset/spankind-manual-span-apis.md
+++ b/.changeset/spankind-manual-span-apis.md
@@ -1,0 +1,5 @@
+---
+'logfire': minor
+---
+
+Expose OpenTelemetry `SpanKind` in the manual span APIs. `span()`, `startSpan()`, `startPendingSpan()`, and `instrument()` accept an optional `kind` that is forwarded to the tracer, and pending span placeholders keep the kind of their real span. Omitting `kind` continues to produce `INTERNAL` spans.
--- a/docs/reference/api.md
+++ b/docs/reference/api.md
@@ -20,6 +20,31 @@ Manual tracing and logging:
 - `reportError(message, error, extraAttributes?, options?)`
 - `Level`
 
+Manual spans accept an optional OpenTelemetry `SpanKind` so remote operations
+(HTTP/RPC/WebSocket clients, server handlers not covered by
+auto-instrumentation, queue producers and consumers) export with an accurate
+kind instead of the `INTERNAL` default:
+
+```ts
+import { SpanKind } from '@opentelemetry/api'
+import * as logfire from 'logfire'
+
+await logfire.span('load live view records', {
+  kind: SpanKind.CLIENT,
+  attributes: {
+    'websocket.endpoint': 'historical_query',
+  },
+  callback: async (span) => {
+    // perform the outbound request/response operation
+  },
+})
+```
+
+`startSpan()`, `startPendingSpan()`, and `instrument()` accept the same `kind`
+option, pending span placeholders keep the kind of their real span, and
+omitting `kind` continues to produce `INTERNAL` spans. Log helpers do not take
+a kind; zero-duration Logfire logs stay `INTERNAL`.
+
 Configuration helpers and utilities:
 
 - `configureLogfireApi()`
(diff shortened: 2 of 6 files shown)


petyosi (changes requested):
There is one runtime contract issue to address before merging in `packages/logfire-api/src/index.ts:442`: `startSpanWithSettings()` now forwards `options.kind` unconditionally, while `logWithSettings()` calls it with `{ ...options, log: true }`. A JavaScript caller—or fully typed TypeScript using `const options: SpanOptions = { kind: SpanKind.CLIENT }`—can therefore pass that object to `info()` and export a `CLIENT` log instead of an `INTERNAL` one. This contradicts the new documentation and changes the previous behavior, where the extra property was ignored. Please only forward `kind` when `resolvedOptions.log !== true` (or strip it on the log path), and add a regression test showing log helpers remain `INTERNAL` when a runtime options object contains `kind`. The intended manual span APIs and documentation otherwise look good.

ayaangazali:
good catch, you're right that the spread through `logWithSettings` let a runtime `kind` leak into logs. fixed in 9bf8f66: `startSpanWithSettings` now drops `kind` when `resolvedOptions.log === true`, plus a regression test where a `SpanOptions`-typed object with `kind: CLIENT` goes through `info()` and the exported span stays `['log', INTERNAL]`. `pnpm run check` green.

ayaangazali:
ready for another look whenever you get a chance: the kind-leak into logs is fixed in 9bf8f66 with a regression test, full check green.

petyosi (approved):

PR #173, opened 2026-07-16
Expose OpenTelemetry SpanKind in manual span APIs

### LINKED ISSUE #171: Expose OpenTelemetry SpanKind in manual span APIs

freddyuniverse:
## Summary

- add a dedicated `SpanOptions` type with optional OpenTelemetry `SpanKind`
- forward `kind` through `span()`, `startSpan()`, `startPendingSpan()`, and `instrument()`
- preserve the requested kind on pending-span placeholders while keeping the default `INTERNAL` behavior
- document a manual remote-operation example and add a `logfire` minor changeset

## Testing

- workspace package build
- `vp run logfire#test` (434 tests passed)
- `vp run logfire#typecheck`
- targeted lint for the changed TypeScript files
- `vp fmt --check`
- `vp check`

Fixes #171

### DIFF
--- a/.changeset/manual-span-kind.md
+++ b/.changeset/manual-span-kind.md
@@ -1,0 +1,6 @@
+---
+'logfire': minor
+---
+
+Allow manual spans and instrumented functions to set OpenTelemetry `SpanKind`, including matching kinds for pending-span placeholders.
+
--- a/docs/packages/logfire.md
+++ b/docs/packages/logfire.md
@@ -130,6 +130,26 @@ await logfire.span('fetch user profile', {
 })
 ```
 
+Set `kind` when a manual span represents a remote operation. This preserves
+OpenTelemetry service topology and dependency semantics:
+
+```ts
+import { SpanKind } from '@opentelemetry/api'
+import * as logfire from 'logfire'
+
+await logfire.span('request inventory service', {
+  kind: SpanKind.CLIENT,
+  attributes: { 'server.address': 'inventory.internal' },
+  callback: async () => fetch('https://inventory.internal/items'),
+})
+```
+
+`span()`, `startSpan()`, `startPendingSpan()`, and `instrument()` accept
+`SpanKind.CLIENT`, `SpanKind.SERVER`, `SpanKind.PRODUCER`, or
+`SpanKind.CONSUMER`. Omitting `kind` keeps OpenTelemetry's default
+`SpanKind.INTERNAL` behavior. Pending placeholders use the same kind as their
+real span.
+
 Use `startSpan()` when you need to control the span lifetime yourself:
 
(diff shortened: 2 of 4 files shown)


freddyuniverse:
I realized #172 was already open for this issue before this PR and now addresses the runtime log-kind edge case raised in review. I’m closing this as a duplicate to avoid splitting the review effort. Sorry I missed the overlap, and thanks for taking a look!

PR #180, opened 2026-07-21
Send logfire-js user agent when exporting traces, logs, and metrics

Kludex:
Identifies the SDK to the Logfire backend on OTLP export requests.

- `@pydantic/logfire-node`: passes `userAgent` to the trace, metric, and log OTLP exporters, resulting in `User-Agent: logfire-js/<version> OTel-OTLP-Exporter-JavaScript/<otel-version>`.
- `@pydantic/logfire-cf-workers`: adds a `User-Agent: logfire-js/<version>` header to the in-process exporter and the tail event export fetch.

The browser package is unchanged because `User-Agent` is a forbidden header in browsers.

## AI Disclaimer

This PR was developed with the assistance of either Claude or Codex. I've reviewed and verified the changes.

### DIFF
--- a/packages/logfire-cf-workers/src/index.ts
+++ b/packages/logfire-cf-workers/src/index.ts
@@ -96,8 +96,9 @@ function getInProcessConfig(config: InProcessConfigOptions): (env: Env) => Trace
       ...config,
       additionalSpanProcessors,
       exporter: {
-        headers: { Authorization: token, 'User-Agent': USER_AGENT },
+        headers: { Authorization: token },
         url: `${baseUrl}/v1/traces`,
+        userAgent: USER_AGENT,
       },
       idGenerator: new ULIDGenerator(),
       postProcessor: (spans: ReadableSpan[]) => postProcessAttributes(spans),
--- a/packages/logfire-cf-workers/src/index.ts
+++ b/packages/logfire-cf-workers/src/index.ts
@@ -30,6 +30,7 @@ import {
 } from 'logfire'
 
 import { exportTailEventsToLogfire } from './exportTailEventsToLogfire'
+import { USER_AGENT } from './userAgent'
 import { LogfireCloudflareConsoleSpanExporter } from './LogfireCloudflareConsoleSpanExporter'
 import { TailWorkerExporter } from './TailWorkerExporter'
 export * from './exportTailEventsToLogfire'
@@ -95,7 +96,7 @@ function getInProcessConfig(config: InProcessConfigOptions): (env: Env) => Trace
       ...config,
       additionalSpanProcessors,
       exporter: {
-        headers: { Authorization: token },
+        headers: { Authorization: token, 'User-Agent': USER_AGENT },
         url: `${baseUrl}/v1/traces`,
       },
       idGenerator: new ULIDGenerator(),
--- a/packages/otel-cf-workers/src/exporter.ts
+++ b/packages/otel-cf-workers/src/exporter.ts
@@ -28,10 +28,20 @@ export class OTLPExporter implements SpanExporter {
   private readonly url: string
   constructor(config: OTLPExporterConfig) {
     this.url = config.url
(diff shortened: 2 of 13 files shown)


packages/logfire-cf-workers/src/index.ts
```diff
@@ -95,7 +96,7 @@ function getInProcessConfig(config: InProcessConfigOptions): (env: Env) => Trace
       ...config,
       additionalSpanProcessors,
       exporter: {
-        headers: { Authorization: token },
+        headers: { Authorization: token, 'User-Agent': USER_AGENT },
```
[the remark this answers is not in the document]
Kludex: @claude do this please

petyosi (changes requested):
Two findings:

1. Please add regression coverage for the new User-Agent behavior across all five export paths: Node traces, metrics, and logs, plus Cloudflare in-process and tail export. Existing Node tests either mock these exporter modules or only exercise processor-constructor compatibility, while the Cloudflare tests only check exports/package metadata. None verifies that `logfire-js/<package version>` reaches exporter configuration or the outbound `fetch` request. This is an externally observable behavior change, and AGENTS.md explicitly requires adding or updating tests for behavior changes. Exact assertions against the package-derived value would also catch a broken `PACKAGE_VERSION` substitution.

2. Non-blocking OTLP specification note: the Node exporters compose the prefix into the recommended `logfire-js/<version> OTel-OTLP-Exporter-JavaScript/<version>` form because the OpenTelemetry `userAgent` option retains the exporter default identifier. The Cloudflare paths send only `logfire-js/<version>`. This is valid HTTP syntax and arguably identifies the JS exporter, so it is not a MUST-level violation, but it does not unambiguously follow the OTLP SHOULD-level recommendation that a distribution identifier retain the default exporter identifier. Consider defining a default identifier for the Cloudflare OTLP exporter and emitting the same two-product form in both Cloudflare paths.

Requesting changes for the missing regression coverage; the specification point is advisory.

Kludex:
@copilot address what petyo said

Kludex:
@petyosi Applied your review: the Cloudflare exporter now has a default `otel-cf-workers/<version>` identifier with the `logfire-js/<version>` prefix prepended, and all five export paths have regression tests with exact assertions. Ready for another look.

packages/otel-cf-workers/src/exporter.ts
```diff
@@ -20,7 +28,11 @@ export class OTLPExporter implements SpanExporter {
   private readonly url: string
   constructor(config: OTLPExporterConfig) {
     this.url = config.url
-    this.headers = { ...defaultHeaders, ...config.headers }
+    this.headers = {
+      ...defaultHeaders,
+      'user-agent': config.userAgent === undefined ? OTLP_EXPORTER_USER_AGENT : `${config.userAgent} ${OTLP_EXPORTER_USER_AGENT}`,
```
petyosi: Please handle `User-Agent` in `config.headers` case-insensitively before this spread. JavaScript object keys are case-sensitive, so an existing caller that passes `{ "User-Agent": "custom/1.0" }` now leaves both that key and the default lowercase `user-agent`; `Headers`/`fetch` normalizes them into `otel-cf-workers/2.0.0, custom/1.0`, which is not the intended product-list form. A lowercase custom `user-agent` instead replaces the exporter identifier entirely.

This is also a compatibility concern because `headers` was the only way to set a custom User-Agent before the new `userAgent` option. Please normalize the configured headers, extract any existing User-Agent case-insensitively, and compose exactly one value with the exporter identifier (with `config.userAgent` taking explicit precedence). Add coverage for both header casings.

Kludex: Done in d70d906 - the constructor now extracts any User-Agent from `config.headers` case-insensitively and composes a single value with the exporter identifier, with `userAgent` taking precedence. Added tests for both casings and the precedence case.

PR #189, opened 2026-07-24
Harden RemoteVariableProvider: reconnect refresh, keepalive backoff, poll jitter, ETag/304, debounced follow-up [murmur:platform/logfire-js-variables-sse-hardening]

dmontagu:
## Summary

Five reliability improvements to `LogfireRemoteVariableProvider`, all internal -- no public API changes, no default behavior changes visible to callers.

### Changes

1. **Reconnect refresh** (`vars/index.ts`): On SSE *re*connects (not the very first connection at startup), immediately call `refresh(true)` to pick up any variable changes that arrived while the stream was down. Controlled by a `sseHadConnected` boolean flag.

2. **Keepalive resets backoff** (`vars/index.ts`): `readSseStream` now returns `true` whenever *any* bytes arrive from the server (including `: keepalive` comment lines), not only on named variable events. This prevents the reconnect delay from doubling to 60 s across healthy-but-quiet stream cycles.

3. **Poll jitter** (`vars/index.ts`): Replaced the fixed `setInterval` with a self-rescheduling `setTimeout` chain. Each interval is `pollingInterval ± 10 %` (uniform). This spreads polling load across instances and avoids thundering-herd on shared infrastructure. `unref()` behavior is preserved.

4. **ETag / conditional GET** (`platform/http.ts`, `vars/index.ts`): Added `requestJsonConditional` to `PlatformAPIClient` (leaves `requestJson` unchanged). When the server returns an `ETag`, subsequent variable fetches send `If-None-Match`; a `304 Not Modified` skips body parsing and config replacement while still advancing `lastFetchedAt`. Servers that do not support ETags see identical behavior.

5. **Debounced follow-up** (`vars/index.ts`): After an SSE-event-triggered `refresh(true)`, schedule one debounced `refresh(true)` ~2 s later. This coalesces rapid bursts of events and survives a short platform-side cache lag between the event and the config endpoint reflecting the change. Timer is cancelled on `shutdown()`.

### Why

These gaps were identified from production reliability analysis. The Python SDK received equivalent fixes in a companion PR. The ETag and caching work pairs with forthcoming platform-side caching/ETag PRs but depends on neither -- the client degrades gracefully when the server omits `ETag` headers.

### Tests

New file `packages/logfire-api/src/vars/remote.test.ts` with 10 test cases using fake timers:

- First SSE connection does not trigger a refresh; subsequent reconnects do
- Keepalive bytes (any bytes) prevent backoff from doubling
- `304` keeps existing config and bumps `lastFetchedAt`; no spurious re-fetch
- `If-None-Match` header is absent on first fetch, present on subsequent fetches
- Jitter is bounded to ±10 % (tested at both extremes via `Math.random` spy)
- `shutdown()` cancels pending polling and debounce timers

[Created via Murmur](https://cloud.murmur.dev/github_app/pydantic/task/github_app%2Fpydantic%2Fw%2Fplatform%2Fgithub_oauth%2Fdmontagu%2Flogfire-js-variables-sse-hardening)

### DIFF
--- a/packages/logfire-api/src/platform/http.ts
+++ b/packages/logfire-api/src/platform/http.ts
@@ -13,6 +13,18 @@ export interface PlatformRequestOptions {
   query?: Record<string, boolean | number | string | undefined>
 }
 
+export interface PlatformConditionalResult {
+  data: unknown
+  etag: string | undefined
+  notModified: false
+}
+
+export interface PlatformNotModifiedResult {
+  notModified: true
+}
+
+export type PlatformConditionalResponse = PlatformConditionalResult | PlatformNotModifiedResult
+
 export class PlatformAPIClient {
   private readonly apiKey: string
   private readonly baseUrl: string
@@ -54,6 +66,57 @@ export class PlatformAPIClient {
         return undefined
       }
       return await parseSuccessJson(response)
+    } catch (error) {
+      if (error instanceof PlatformConfigurationError || error instanceof PlatformHTTPError || error instanceof PlatformJSONError) {
+        throw error
+      }
+      if (isAbortError(error)) {
+        throw new PlatformTimeoutError('Logfire platform API request timed out')
+      }
+      if (error instanceof PlatformTransportError) {
+        throw error
+      }
+      throw new PlatformTransportError(`Logfire platform API request failed: ${formatError(error)}`)
+    } finally {
+      clearTimeout(timeout)
+    }
+  }
+
+  async requestJsonConditional(
+    path: string,
+    options: PlatformRequestOptions & { ifNoneMatch?: string } = {}
+  ): Promise<PlatformConditionalResponse> {
+    const controller = new AbortController()
+    const timeout = setTimeout(() => {
+      controller.abort()
+    }, this.timeoutMs)
+
+    try {
+      const headers = this.buildHeaders()
+      if (options.ifNoneMatch !== undefined) {
+        headers.set('If-None-Match', options.ifNoneMatch)
+      }
+      const init: RequestInit = {
+        headers,
+        method: options.method ?? (options.body === undefined ? 'GET' : 'POST'),
+        signal: controller.signal,
+      }
+      if (options.body !== undefined) {
+        init.body = JSON.stringify(options.body)
+      }
+      const response = await this.fetchImpl(this.buildUrl(path, options.query), init)
+      if (response.status === 304) {
+        return { notModified: true }
+      }
+      if (!response.ok) {
+        throw new PlatformHTTPError(response.status, response.statusText, await parseErrorDetail(response))
+      }
+      if (response.status === 204) {
+        return { data: undefined, etag: undefined, notModified: false }
+      }
+      const data = await parseSuccessJson(response)
+      const etag = response.headers.get('etag') ?? undefined
+      return { data, etag, notModified: false }
     } catch (error) {
       if (error instanceof PlatformConfigurationError || error instanceof PlatformHTTPError || error instanceof PlatformJSONError) {
         throw error
--- a/packages/logfire-api/src/platform/http.ts
+++ b/packages/logfire-api/src/platform/http.ts
@@ -42,44 +42,9 @@ export class PlatformAPIClient {
   }
 
   async requestJson(path: string, options: PlatformRequestOptions = {}): Promise<unknown> {
-    const controller = new AbortController()
-    const timeout = setTimeout(() => {
-      controller.abort()
-    }, this.timeoutMs)
-
-    try {
-      const init: RequestInit = {
-        headers: this.buildHeaders(),
-        method: options.method ?? (options.body === undefined ? 'GET' : 'POST'),
-        signal: controller.signal,
-      }
-      if (options.body !== undefined) {
-        init.body = JSON.stringify(options.body)
-      }
-      const response = await this.fetchImpl(this.buildUrl(path, options.query), {
-        ...init,
-      })
-      if (!response.ok) {
-        throw new PlatformHTTPError(response.status, response.statusText, await parseErrorDetail(response))
-      }
-      if (response.status === 204) {
-        return undefined
-      }
-      return await parseSuccessJson(response)
-    } catch (error) {
-      if (error instanceof PlatformConfigurationError || error instanceof PlatformHTTPError || error instanceof PlatformJSONError) {
-        throw error
-      }
-      if (isAbortError(error)) {
-        throw new PlatformTimeoutError('Logfire platform API request timed out')
-      }
-      if (error instanceof PlatformTransportError) {
-        throw error
-      }
-      throw new PlatformTransportError(`Logfire platform API request failed: ${formatError(error)}`)
-    } finally {
-      clearTimeout(timeout)
-    }
+    const result = await this.requestJsonConditional(path, options)
+    // notModified is only possible when ifNoneMatch is passed; requestJson never sends it.
+    return result.notModified ? undefined : result.data
   }
 
   async requestJsonConditional(
--- a/packages/logfire-api/src/vars/index.ts
+++ b/packages/logfire-api/src/vars/index.ts
@@ -522,7 +522,7 @@ export class LogfireRemoteVariableProvider implements VariableProvider {
     }
 
     const now = Date.now()
-    if (!force && this.lastFetchedAt !== undefined && now - this.lastFetchedAt < this.pollingIntervalMs) {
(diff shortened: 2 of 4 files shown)


petyosi (approved):
Reviewed the SSE reconnect/backoff changes, jittered polling lifecycle, conditional ETag/304 handling, debounce behavior, and shutdown paths. I found no actionable issues. `pnpm run check` passes on the PR head.

packages/logfire-api/src/vars/index.ts
```diff
@@ -720,14 +749,17 @@ export class LogfireRemoteVariableProvider implements VariableProvider {
...
+    let receivedAnyContent = false
     try {
       while (!this.shutdownRequested) {
         // eslint-disable-next-line no-await-in-loop -- stream chunks must be read sequentially.
         const { done, value } = await reader.read()
         if (done) {
           break
         }
+        if (value.byteLength > 0) {
+          receivedAnyContent = true
```
petyosi: Resetting backoff on *any* bytes can defeat the exponential backoff entirely in one failure mode: a misbehaving proxy/LB that returns HTTP 200 with a short non-SSE body (e.g., an HTML error page) that closes immediately. Each cycle delivers bytes, so the delay resets to 1s every time — a permanent 1 req/s reconnect loop where the old code would back off to 60s.

Two cheap ways to keep the keepalive fix while preserving backoff against broken endpoints: only set `receivedAnyContent` when a line parses as SSE framing (starts with `:` or a `field:` prefix), or only reset the delay when the connection stayed open longer than some minimum duration.

packages/logfire-api/src/vars/index.ts
```diff
@@ -472,13 +475,21 @@ export class LogfireRemoteVariableProvider implements VariableProvider {
...
-      const maybeTimer = this.pollingTimer as { unref?: () => void }
-      if (typeof maybeTimer.unref === 'function') {
-        maybeTimer.unref()
+      const scheduleNextPoll = (): void => {
+        if (this.shutdownRequested) {
+          return
+        }
+        const jitter = (Math.random() * 0.2 - 0.1) * this.pollingIntervalMs
+        this.pollingTimer = setTimeout(() => {
+          this.refresh().catch(ignoreBackgroundError)
```
petyosi, line 484: Negative jitter interacts with the freshness check in `refresh()` (`now - lastFetchedAt < this.pollingIntervalMs` → early return): a poll that fires at interval − 10% is always a no-op, so the effective fetch cadence drifts toward ~1.5–2× the configured interval.

This is partially pre-existing — the old `setInterval` code stamped `lastFetchedAt` at fetch *completion*, so the next tick also landed just inside the freshness window — but since this PR is about polling reliability it may be worth fixing here: compare against `this.pollingIntervalMs * 0.9` (interval minus max jitter) in the freshness check.

packages/logfire-api/src/platform/http.ts
```diff
@@ -70,6 +82,57 @@ export class PlatformAPIClient {
     }
   }
 
+  async requestJsonConditional(
```
petyosi, line 50: This duplicates ~40 lines of `requestJson` verbatim (timeout setup, method inference, body serialization, and the four-branch error-mapping `catch`). Future error-handling fixes would need to be made twice.

`requestJson` could delegate here — `notModified` is impossible without `ifNoneMatch`, so it can just unwrap `result.data` — or both could share an extracted core.

packages/logfire-api/src/vars/index.ts
```diff
@@ -512,9 +527,17 @@ export class LogfireRemoteVariableProvider implements VariableProvider {
...
     this.refreshPromise = this.transport
-      .requestJson('/v1/variables/', { method: 'GET' })
-      .then((data) => {
-        this.config = normalizeVariablesConfig(data)
+      .requestJsonConditional('/v1/variables/', {
+        method: 'GET',
+        ...(this.lastEtag !== undefined ? { ifNoneMatch: this.lastEtag } : {}),
+      })
+      .then((result) => {
+        if (result.notModified) {
```
petyosi: Nit: inverting to `if (!result.notModified) { ... }` with the comment moved above the shared `this.lastFetchedAt = Date.now()` assignment avoids the empty branch.

petyosi (commented):
Verified locally: all 466 `logfire` package tests pass (including the 10 new ones) and typecheck is clean.

The implementation matches the PR description, reuses the existing `refreshPromise`/`queuedForcedRefreshPromise` dedup machinery correctly, and `unref()`/shutdown handling is consistent for all new timers. The 304 handling is careful: `lastFetchedAt` advances without touching config, and a later 200 without an `ETag` correctly clears `lastEtag`.

A few non-blocking comments inline. One repo-convention item not tied to a line: these are behavior changes shipped in the published `logfire` package (new `If-None-Match` headers, extra refreshes on reconnect/debounce), so a patch changeset with a short note seems warranted even though the API is unchanged.

packages/logfire-api/src/vars/index.ts
dmontagu: Fixed -- inverted to `if (!result.notModified)` with the comment moved above the shared `lastFetchedAt` assignment. No empty branch remains.

packages/logfire-api/src/vars/index.ts
dmontagu: Good catch. Went with the SSE-framing approach: `receivedAnyContent` is now set only when a completed line starts with `:` (comments, e.g. `: keepalive`) or one of the named SSE fields (`data:`, `event:`, `id:`, `retry:`). An HTML error page won't match any of those patterns, so a short misbehaving-proxy body won't reset the backoff. The per-connection-age approach would also work but is harder to test deterministically with fake timers.

packages/logfire-api/src/vars/index.ts
dmontagu: Fixed -- freshness check now uses `pollingIntervalMs * 0.9` so polls that fire at the minimum-jitter delay (-10%) are not silently dropped.

packages/logfire-api/src/platform/http.ts
dmontagu: Fixed -- `requestJson` now delegates to `requestJsonConditional` in two lines. The `notModified` branch is unreachable without `ifNoneMatch`, so it just falls through to `result.data`.

packages/logfire-api/src/vars/remote.test.ts
```diff
@@ -0,0 +1,497 @@
...
+    const varFetchesAfterFirstConnect = urls.filter((u) => u.includes('/v1/variables/')).length
+    // First connection must NOT trigger a variables fetch.
+    expect(varFetchesAfterFirstConnect).toBe(0)
+
+    // Advance past the 1000ms reconnect delay so the second connection fires.
+    await vi.advanceTimersByTimeAsync(1100)
+
+    const varFetchesAfterReconnect = urls.filter((u) => u.includes('/v1/variables/')).length
+    // Second connection (reconnect) MUST trigger a variables fetch.
+    expect(varFetchesAfterReconnect).toBeGreaterThanOrEqual(1)
```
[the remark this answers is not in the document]
dmontagu: Fixed -- changed to `toBe(1)`.

packages/logfire-api/src/vars/index.ts
dmontagu: Good catch — took the SSE-framing option. `readSseStream` now only sets `receivedAnyContent` for lines that are actually SSE-framed (a `:` comment or a `data:`/`event:`/`id:`/`retry:` field), so an HTML error page delivers bytes without resetting the delay and the backoff still climbs to 60s.

This applied to the Python SDK too, which had the same reset-on-any-line logic — fixed there in pydantic/logfire#2144 with a matching test, so the two SDKs behave the same.

packages/logfire-api/src/vars/index.ts
dmontagu: Fixed as suggested — the freshness check in `refresh()` now compares against `this.pollingIntervalMs * 0.9`, so a poll landing at interval − 10% is no longer swallowed and the effective cadence tracks the configured interval.

packages/logfire-api/src/platform/http.ts
dmontagu: Deduplicated — `requestJson` now delegates to `requestJsonConditional` and just unwraps `result.data` (`notModified` is unreachable without `ifNoneMatch`). The timeout, method inference, body serialization and error mapping all live in one place again.

packages/logfire-api/src/vars/index.ts
dmontagu: Done — inverted to `if (!result.notModified)` with the comment above the shared `lastFetchedAt` assignment, so there is no empty branch.

PR #194, opened 2026-07-29
docs: rework TypeScript Getting Started (site-standard voice + automatic instrumentation)

strawgate:
Reworks the TypeScript **Getting Started** so it matches the flow, voice, and quality of the rest of the Logfire docs (which all render together on pydantic.dev/docs), while staying idiomatic and lovable for TypeScript developers.

## What changed
- **Before you start** — create an account and project, and see where the write token comes from, instead of being dropped straight into `npm install`.
- **Explains "span"** on first use and links Core Concepts.
- **Get automatic traces** — the core OpenTelemetry value that was missing: install the auto-instrumentations and run with `node --import`, so you get HTTP, database, and framework traces without writing spans by hand. Points to the Node.js package and framework guides.
- **See it in the Live view** (a real verify step) and a **Troubleshooting** table for the common TS/OTel pitfalls (missing credentials, flush-before-exit, `--import` ordering, Node version).
- **Next steps** fan-out, plus an optional copy-paste **AI-agent setup** prompt.
- Sentence-case headings and an aligned Overview, matching the site's voice.

TypeScript touches: ESM (`.mjs`), `npx tsx` for `.ts` files, `node --import`, npm/pnpm, and the frameworks TS devs actually use.

All cross-repo links use full `pydantic.dev/docs/...` URLs and resolve (200); relative links verified; no em-dashes.

## Follow-up (not in this PR)
The deeper TS pages (packages, frameworks, reference) are still Title Case. Standardizing their headings/voice to match is a separate sweep.

### DIFF
--- a/docs/get-started.md
+++ b/docs/get-started.md
@@ -1,73 +1,127 @@
 ---
-title: Getting Started
-description: Install the Logfire TypeScript SDK, configure a write token, and send your first span from Node.js.
+title: Getting started
+description: Install the Logfire TypeScript SDK, send your first span from Node.js, and get automatic traces from the frameworks you already use.
 ---
 
-# Getting Started
+# Getting started
 
-The fastest way to try the TypeScript SDK is a small Node.js script. You need a Logfire write token, Node.js 24 or newer, and a package manager such as npm or pnpm.
+Go from install to your first trace in about five minutes. A **trace** is the full record of one request, job, or task from start to finish, made of nested **spans**; a span is one operation within it, with a name, a start, and a duration. The Logfire TypeScript SDK is built on [OpenTelemetry](https://opentelemetry.io/), the open industry standard, and works across Node.js, browsers, Cloudflare Workers, and other OpenTelemetry-compatible runtimes.
 
-## Install
+## Before you start
 
-Create a project and install the Node package:
+You need a Logfire project to send data to, and a recent Node.js:
+
+1. [Create a free account](https://logfire.pydantic.dev/login) and create a project when prompted. A project is a namespace that holds your data.
+2. Install Node.js 24 or newer, with a package manager such as npm or pnpm.
+
+> Your telemetry is stored in the Logfire [data region](https://pydantic.dev/docs/logfire/manage/data-regions/) you choose when you create your account.
+
+## Let an AI agent set it up
+
+To have an AI coding agent wire this up for you, paste this into Claude Code, Cursor, or a similar tool:
+
+```text
+Set up Pydantic Logfire in this TypeScript project. Read
+https://pydantic.dev/docs/logfire/typescript-sdk/get-started/ and follow it: install
+@pydantic/logfire-node, call logfire.configure() in an instrumentation.ts file loaded before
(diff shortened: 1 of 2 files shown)


strawgate:
Good catch — you're right that `node hello.ts` needs an ES-module project. Fixed: `tsx` is now the primary runner (it handles TypeScript and ESM in any project), and the note about running with `node` directly now says the project must be ESM (`"type": "module"`). Troubleshooting row updated too.

docs/get-started.md
```diff
@@ -1,73 +1,127 @@
...
-description: Install the Logfire TypeScript SDK, configure a write token, and send your first span from Node.js.
+title: Getting started
+description: Install the Logfire TypeScript SDK, send your first span from Node.js, and get automatic traces from the frameworks you already use.
 ---
 
-# Getting Started
+# Getting started
 
-The fastest way to try the TypeScript SDK is a small Node.js script. You need a Logfire write token, Node.js 24 or newer, and a package manager such as npm or pnpm.
+Go from install to your first trace in about five minutes. A **trace** is the journey of one request, made of nested **spans**; a span is one operation, with a name, a start, and a duration. The Logfire TypeScript SDK is built on [OpenTelemetry](https://opentelemetry.io/), the open industry standard, and works across Node.js, browsers, Cloudflare Workers, and other OpenTelemetry-compatible runtimes.
```
petyosi: Framing the trace as "the journey of one request" sounds strange. Can we get a more accurate, less abstract wording for this? I might get this wrong, but traces don't always equal requests.

docs/get-started.md
```diff
@@ -1,73 +1,127 @@
...
+npx tsx hello.ts
+```
+
+> These examples are ES modules. `tsx` runs them in any project. To run them with `node` directly (Node.js 24 strips the types), your project needs to be an ES module: add `"type": "module"` to your `package.json`.
+
+## See it in the Live view
+
+Open the [Live view](https://pydantic.dev/docs/logfire/observe/live/) for your project. Your `greeting` trace appears as it arrives: one row for the span, with the `Hello world!` log nested inside it. Click the span to open its full trace and read its details.
+
+## Get automatic traces
```
petyosi, line 84: Just double checking - this is a genuinely new content, and it actually recommends the auto instrumentations approach - some auto instrumentations are useful, other less so.

docs/get-started.md
strawgate: Good point — reworded to "the full record of one request, job, or task from start to finish" so it no longer implies a trace is always a request. (Heads up: the Python "Send your first trace" page has the same "journey of one request" phrasing, so that's worth a follow-up fix there too.)

docs/get-started.md
strawgate: Fair. I kept the recommendation since `auto-instrumentations-node` is the standard OpenTelemetry on-ramp, but added the nuance you're after: the note now says it turns them all on at once, that a few (like the filesystem instrumentation) are noisy, and links the Node.js package to enable/disable individual ones.

strawgate:
Fixed in #195 — the `hello.ts` example used top-level `await`, which fails under `tsx` in a CommonJS project. Wrapped it in an async `main()` so it runs in any project.

PR #206, opened 2026-08-03
Record exceptions from rejected zone.js promises in span callbacks

ayaangazali:
In `spanWithSettings` (`packages/logfire-api/src/index.ts`) a callback returning a native promise gets `recordSpanException` on rejection, which records the error and sets the span status to `ERROR`. The zone.js branch right below it, the one the "clunky detection" comment refers to, only attaches `.finally(() => span.end())`. So under zone.js, which is to say in Angular apps, a span whose callback rejects ends up looking successful: no exception event and no error status, even though the rejection still propagates to the caller.

This adds the rejection handler to that branch when the value also exposes `then`, reusing the same `recordSpanException` call as the native path. The returned chain is ignored on purpose, exactly as the native branch does, so `result` is still handed back untouched and the caller keeps seeing the rejection. I deliberately kept this scoped to the `.finally` branch rather than treating every thenable as a promise, because the existing "thenable callback result is returned untouched" test pins the behavior for values that only have `then`.

The new test builds a non-native object exposing `then` and `finally` the way a zone.js promise does, and it fails on current main with zero calls to `recordException`. Ran the repo preflight and `pnpm run check`, both green. Patch changeset for `logfire` included.

quick disclosure: worked through this with Claude Code and reviewed the diff myself. Freshman here, and I do not have a real Angular app to hand, so if the zone.js shape I mocked is off I would be glad to be corrected :)

### DIFF
--- a/.changeset/thenable-exception-recording.md
+++ b/.changeset/thenable-exception-recording.md
@@ -1,0 +1,5 @@
+---
+'logfire': patch
+---
+
+Record the exception on the span when a zone.js style promise returned from a `span()` callback rejects. Native promise rejections already recorded the error and set the span status to `ERROR`, but the zone.js branch only ended the span, so in Angular applications a failing span looked successful and carried no exception.
--- a/packages/logfire-api/src/index.test.ts
+++ b/packages/logfire-api/src/index.test.ts
@@ -837,35 +837,42 @@ describe('span', () => {
 
   test('records the exception when a zone.js style promise rejects', async () => {
     const error = new Error('zone-oops')
-    // Not a native Promise, but exposes `then` and `finally` the way a zone.js promise does.
-    const settled = Promise.reject(error).catch(() => undefined)
-    const zonePromise = {
-      finally: (onFinally: () => void) => {
-        settled.then(onFinally, onFinally)
-        return zonePromise
-      },
-      then: (_onFulfilled: undefined, onRejected: (reason: unknown) => void) => {
-        settled.then(
-          () => {
-            onRejected(error)
-          },
-          () => {
-            onRejected(error)
-          }
-        )
-        return zonePromise
-      },
+    // zone.js patches the global Promise, so an intrinsic promise from a native async function
+    // fails `instanceof Promise` while still exposing `then` and `finally`.
+    // Typed as a bare thenable, not a Promise, so the callback is not treated as async.
+    type ZonePromiseLike = { finally: (onFinally: () => void) => unknown; then: (...args: never[]) => unknown }
+    const intrinsic = Promise.reject(error) as unknown as ZonePromiseLike
+    const unhandled: unknown[] = []
+    const onUnhandled = (reason: unknown) => {
+      unhandled.push(reason)
+    }
+    process.on('unhandledRejection', onUnhandled)
+    const OriginalPromise = globalThis.Promise
+    const ZoneAwarePromise = function ZoneAwarePromise() {
+      // Stands in for zone.js replacing the global Promise binding.
+    }
+    globalThis.Promise = ZoneAwarePromise as unknown as PromiseConstructor
+
+    let result: unknown
+    try {
+      result = span('test', { callback: () => intrinsic })
+    } finally {
+      globalThis.Promise = OriginalPromise
     }
 
-    const result = span('test', { callback: () => zonePromise })
-
-    expect(result).toBe(zonePromise)
-    await settled
-    await Promise.resolve()
+    expect(result).toBe(intrinsic)
+    // The original rejection still reaches the caller.
+    await expect(intrinsic as unknown as Promise<never>).rejects.toBe(error)
+    await new OriginalPromise<void>((resolve) => {
+      setTimeout(resolve, 0)
+    })
 
     expect(spanMock.recordException).toHaveBeenCalledWith(error)
     expect(spanMock.setStatus).toHaveBeenCalledWith({ code: SpanStatusCode.ERROR, message: 'Error: zone-oops' })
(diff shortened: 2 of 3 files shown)


petyosi (changes requested):
This fixes a real telemetry gap, but I think the fallback should use one settlement subscription before we merge it.

Calling both `.then(undefined, onRejected)` and `.finally(onFinally)` creates two derived chains. The ignored promise returned by `.finally()` preserves the rejection, so it can report an unhandled rejection even when the caller handles the original promise. This behavior already exists in the fallback, but this change should avoid carrying it forward.

Please use a single `.then(onFulfilled, onRejected)` subscription for promise-like values in this branch:

- end the span in the fulfillment handler;
- record the exception and end the span in the rejection handler;
- continue to return the original result unchanged.

Keeping the existing `finally` check as the gate is reasonable if we want to avoid consuming arbitrary lazy thenables.

Please also make the regression test representative of the zone.js case. With zone.js patching global `Promise`, a native `async` function can return an intrinsic native promise for which `result instanceof Promise` is false, while `finally` is present. The test should verify that the original rejection reaches the caller, the exception and error status are recorded, the span ends once, and no extra unhandled rejection is produced.

ayaangazali:
Good catch on the second chain, and you are right that it was mine to avoid rather than inherit. 4a96d77 switches to a single `then(onFulfilled, onRejected)` subscription: the fulfillment handler ends the span, the rejection handler records the exception and ends the span, and `result` is still returned untouched. The `finally` check stays as the gate so lazy thenables are left alone, and `finally` is now only called when `then` is absent.

The test is rebuilt around the shape you described. It replaces the global `Promise` with a stand-in so an intrinsic promise fails `instanceof Promise` while keeping `then` and `finally`, then asserts the original rejection still reaches the caller, the exception and `ERROR` status are recorded, the span ends exactly once, and an `unhandledRejection` listener stays empty.

I checked it fails for both regressions rather than just passing: against main it fails with zero `recordException` calls, and against the previous two-chain version it fails on the empty-unhandled assertion, which is the exact problem you flagged. Ran the repo preflight and `pnpm run check`, both green.

ayaangazali:
Flagging for re-review, since this still shows as changes requested but the change you asked for landed on 2026-08-04.

The two derived chains are gone. Settlement is observed through a single `then` subscription, so the chain returned by `finally` is no longer left rejected and unhandled when the caller handles the original, and `result` is still returned untouched.

Rebased onto current main (817fca1) and re-verified there, since the original run predated the dependency refresh in #212. Build is green, and the mutation check still holds: dropping the rejection handler fails the zone.js test with `expected "vi.fn()" to be called with arguments: [ Error: zone-oops ]` and nothing else.

petyosi (changes requested):
One correctness regression remains in the new settlement path.

`packages/logfire-api/src/index.ts:622`: The new `then` lookup and the call below are both unguarded. A callback result with a working `finally` but a throwing `then` getter worked on `main`, which only called `finally`; this branch now makes `span()` throw after the callback succeeded and never ends the span. A `then()` method that throws synchronously has the same result.

Please protect both lookup and subscription, preserve the original result, ensure the span ends, and add regressions for both cases.

ayaangazali:
You were right and my previous comment answered the wrong review. The 08-04 single-settlement change had landed, but the 08-10 point about the unguarded lookup was still open on the branch, and I read it as already covered. Fixed now in 8acb8e8.

`then` is read once into a local and only that reference is called, so a getter that returns a function on one read and throws on the next cannot reach the call. The two failure modes get different fallbacks, since they are not equivalent:

- **Lookup throws.** Nothing has been called yet, so it falls through to `finally`, which is exactly what `main` did for such a value. Ending the span immediately here would have replaced the leak with a wrong duration.
- **The call throws.** `then` may already have registered the handlers, so adding a `finally` subscription risks ending the span twice, and doing nothing risks never ending it. It ends the span there, behind a flag so a handler that already fired cannot end it again.

Two regressions, each verified by reverting the source:

- throwing `then` getter with a working `finally`: the result is returned unchanged, the span is still open, and it ends once when `finally` settles. On the previous commit this fails with `then getter boom`.
- `then` that invokes its fulfillment handler and then throws: the span ends exactly once. Dropping the flag alone fails this with two calls to `end`.

I left the `finally` call itself unwrapped. It is already known to be a function because it gates the branch, and a `finally` that throws when called behaved the same way on `main`, so guarding it would be handling a state this change does not introduce.

Green on 79bcac0. Happy to close this if the zone.js path is not worth carrying.

ayaangazali:
Rebased onto `8effa79` and simplified against what just landed.

#236 merged `getThenMethod`, which does exactly what this branch was hand-rolling in the zone.js path: guard the `then` read, and hand back the method so the call site cannot re-read it. This now uses that helper instead, which drops 18 lines and leaves one implementation of the rule in the file rather than two. `ThenMethod` gained an optional `onRejected` parameter, since this branch needs both handlers and that is the real shape of `then`. The `resultWithFinally` cast narrowed to just `finally`, which is all it gates on now.

Behaviour is unchanged from your last review. Both regressions still bite: replacing `getThenMethod` with a raw property read fails the throwing-getter test, and dropping the `ended` flag fails the ends-once test. 495 tests green on the new base, preflight and `pnpm run check` clean.

petyosi:
Merged locally and landed on main as eca8805 — thanks for the thorough follow-through across the review rounds; both sets of requested changes were verified as addressed. Two follow-ups landed alongside it: 03b4d62 moves the test's unhandledRejection cleanup into a finally block, and cd89805 records the failure on the span when then throws before any settlement handler runs, so that path no longer closes with an OK status. (Squash-merged outside GitHub, so closing manually.)

PR #207, opened 2026-08-03
Mark D1 batch statement spans as errors when the batch fails

ayaangazali:
`instrumentD1Fn` creates one span per statement in a `batch()` call (`packages/otel-cf-workers/src/instrumentation/d1.ts:116`), but the catch only sets `ERROR` on the parent span. The statement spans are ended in the `finally` with their status still `UNSET`, so a failed batch renders as a failed parent with a set of apparently successful queries underneath it, which reads exactly backwards from what happened.

This sets the error status on the statement spans too. I left the exception itself on the parent rather than recording it once per statement, since it is one failure rather than N, but say the word if you would rather each statement span carried it.

There was no test file for the D1 bindings, so I added one covering a successful batch and a rejected one. The rejection test fails on current main with `expected +0 to be 2`, that being `UNSET` where `ERROR` was expected. Ran the repo preflight and `pnpm run check`, both green. Patch changeset for `@pydantic/otel-cf-workers` included.

Worth noting this one hid from the sweep I used for #201, #203 and #204: d1.ts pairs `recordException` and `setStatus` three times each, so it looked balanced, and only reading the batch branch showed the sub-spans were missed.

quick disclosure: built with Claude Code's help and I read the final diff myself. Freshman still learning, so corrections welcome :)

### DIFF
--- a/.changeset/d1-batch-subspan-status.md
+++ b/.changeset/d1-batch-subspan-status.md
@@ -1,0 +1,5 @@
+---
+'@pydantic/otel-cf-workers': patch
+---
+
+Mark the per-statement spans of a D1 batch as errors when the batch fails. `instrumentD1Fn` created one span per query in a `batch()` call but only set `ERROR` on the parent span, so a failed batch showed every query as apparently successful underneath a failed parent. The statement spans now carry the error status too, while the exception itself stays on the parent so it is not duplicated once per statement.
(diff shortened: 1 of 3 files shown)


petyosi (approved):

ayaangazali:
Rebased onto current main (79bcac0) and re-verified there. Approved and clean, no conflicts.

Recap since it has been a couple of weeks: `instrumentD1Fn` opens one sub-span per statement in a batch, but the catch sets ERROR only on the parent, so a failed batch rendered as a failed parent over N apparently successful queries. The rejection test fails on main with `expected +0 to be 2`.

Happy to close it if you would rather not take this one.

ayaangazali:

PR #209, opened 2026-08-04
Record http.url on cache spans for string and URL keys

ayaangazali:
The Cache methods accept `RequestInfo | URL`, so the key can be a `Request`, a bare string, or a `URL`. `instrumentFunction` in `packages/otel-cf-workers/src/instrumentation/cache.ts` read the url as `argArray[0].url`, which only exists on a `Request`, so `caches.default.match('https://example.com/x')` and the `URL` form both produced a cache span with no `http.url` at all. Two of the three documented input shapes lost the attribute.

This pulls the url out of whichever shape was passed. I also made `sanitiseURL` return undefined instead of throwing on something unparseable, because it runs while building the span attributes, before the span exists, so a throw there would have surfaced inside the caller's own cache call rather than as a failed span.

There was no test file for the cache bindings, so I added one covering all three key shapes plus an unparseable one. The string and URL cases fail on current main with `expected undefined to be 'https://example.com/...'`. Ran the repo preflight and `pnpm run check`, both green. Patch changeset for `@pydantic/otel-cf-workers` included.

quick disclosure: worked through this with Claude Code and read the final diff myself. Freshman still learning, so if you would rather keep the attribute Request-only I am happy to drop it :)

### DIFF
--- a/.changeset/cache-url-attribute.md
+++ b/.changeset/cache-url-attribute.md
@@ -1,0 +1,5 @@
+---
+'@pydantic/otel-cf-workers': patch
+---
+
+Record `http.url` on Cache spans when the key is a string or a `URL`, not just a `Request`. The Cache methods accept `RequestInfo | URL`, but the attribute was read from `argArray[0].url`, which only exists on a `Request`, so `caches.default.match('https://example.com/x')` produced a span with no URL at all. Malformed keys now omit the attribute instead of letting URL parsing throw into the caller.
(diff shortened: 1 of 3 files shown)


petyosi (approved):

ayaangazali:
Rebased onto current main (79bcac0) and re-verified there. Approved and clean, no conflicts.

Recap: Cache methods accept `RequestInfo | URL`, but the attribute was read as `argArray[0].url`, which only exists on a `Request`. String keys and `URL` keys both produced cache spans with no `http.url`, so two of the three documented input shapes lost it.

Happy to close it if you would rather not take this one.

ayaangazali:

PR #214, opened 2026-08-06
Stop onStart throwing on a non-absolute http.url

ayaangazali:
`LogfireSpanProcessor.onStart` renames fetch spans by parsing the `http.url` attribute with `new URL(...)`, with nothing around it (`packages/logfire-browser/src/LogfireSpanProcessor.ts`). `onStart` runs for every span, so an `http.url` that is not an absolute URL raises `TypeError: Invalid URL` out of the processor and into the application's own span creation. A relative value like `/api/thing` or a non-string value both do it. That makes instrumentation break user code, which is the wrong direction for a tracing SDK.

The parse now returns undefined instead of throwing and the span name is left as it was, which follows the same shape as `resolveUrl` in `telemetryUrls.ts` where URL parsing is already guarded this way.

There was no test file for this processor, so I added one covering the fetch rename, the interaction rename, and both unparseable cases. The two unparseable tests fail on current main with `TypeError: Invalid URL`. Ran the repo preflight and `pnpm run check`, both green. Patch changeset for `@pydantic/logfire-browser` included.

For what it is worth I did not find a first-party instrumentation that emits a relative `http.url` today, so this is defence in depth rather than a bug someone is currently hitting. It seemed worth guarding anyway given the blast radius if it ever happens.

quick disclosure: built with Claude Code's help and I read the diff myself. Freshman still learning, so push back if you would rather leave the parse unguarded :)

### DIFF
--- a/.changeset/span-processor-url-guard.md
+++ b/.changeset/span-processor-url-guard.md
@@ -1,0 +1,5 @@
+---
+'@pydantic/logfire-browser': patch
+---
+
+Stop `LogfireSpanProcessor.onStart` throwing when a span carries an `http.url` that is not an absolute URL. The processor parsed the attribute with `new URL()` while renaming fetch spans, so a relative or non-string value raised `TypeError: Invalid URL` out of `onStart` and broke span creation for the application. Unparseable values now leave the span name unchanged.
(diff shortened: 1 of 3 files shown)


petyosi (approved):

PR #216, opened 2026-08-07
Expand references that follow an even backslash run

ayaangazali:
## What is wrong

`expandReferenceSerializedValue` guards nested composition with a regex over the **JSON-encoded** value:

```ts
export const HAS_REFERENCE: RegExp = /(?<!\\)@\{/u
...
if (!HAS_REFERENCE.test(serializedValue)) {
  return { composedFrom: [], serializedValue, value: parsed }
}
```

Escape state is a property of the decoded string, and the rest of the module agrees: `isEscapedAt()` counts backslash **parity**, so an even run does not escape the reference. The lookbehind only checks the single preceding character, and JSON doubles every backslash, so a value whose reference is preceded by an even (non-escaping) run reaches the guard looking escaped and skips composition entirely.

## Evidence

The same string expands at the top level but not when reached through a reference:

```ts
// through a reference
await expandReferences(
  JSON.stringify('@{ref}@'),
  resolver({ inner: resolved('X'), ref: resolved('\\\\@{inner}@') })
)
// serializedValue -> "\\\\@{inner}@"   (inner never expands, absent from composedFrom)

// the same value at the top level
await expandReferences(JSON.stringify('\\\\@{inner}@'), resolver({ inner: resolved('X') }))
// serializedValue -> "\\X"             (expands)
```

Mapping literal backslashes before a reference to what the guard decides:

| literal backslashes | escaped per `isEscapedAt` | guard expands |
|---|---|---|
| 0 | no | yes |
| 1 | yes | no |
| 2 | no | **no** |
| 3 | yes | no |

Only the even-and-nonzero rows are wrong, so this is the single case the change affects.

## What this does

Uses the existing `hasReferences()` helper on the already-parsed value. It walks strings, arrays and objects and calls `hasCompositionReferences()`, which is the same parity rule `isEscapedAt()` implements, so the guard and the expander now agree.

Swapping the regex for `hasCompositionReferences(serializedValue)` would be wrong in the other direction: parity applied to JSON text treats a single literal backslash as even and would break the escaped case pinned at `composition.test.ts:172`. Checking the decoded value is what keeps both branches correct.

`HAS_REFERENCE` had this one caller and is not re-exported from `vars/reference-syntax`, so it is removed rather than left as a second, disagreeing definition.

Tests cover both branches: even run expands and matches the top-level result, odd run stays literal. Verified by reverting the guard, which fails exactly these two tests and no others.

---
This change was prepared with AI assistance and reviewed by me before opening.

### DIFF
--- a/.changeset/composition-escaped-backslash-guard.md
+++ b/.changeset/composition-escaped-backslash-guard.md
@@ -1,0 +1,5 @@
+---
+'logfire': patch
+---
+
+Expand references in a referenced variable's value when they follow an even backslash run. The fast-path guard in `expandReferenceSerializedValue` tested the JSON-encoded text with a single-character lookbehind, so any `@{ref}@` preceded by a literal backslash was treated as escaped. A value such as `\\@{inner}@` therefore skipped composition entirely and `inner` was left unexpanded and absent from `composedFrom`, even though the same string expands when it appears at the top level.
--- a/packages/logfire-api/src/vars/composition.test.ts
+++ b/packages/logfire-api/src/vars/composition.test.ts
@@ -174,6 +174,35 @@ describe('variable composition', () => {
     })
   })
 
+  it('applies the backslash escape rule to referenced values the same way as to top-level values', async () => {
+    // An even backslash run leaves the reference live, so it expands on both paths.
+    await expect(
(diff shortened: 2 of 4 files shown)


ayaangazali:
Tightened the escape-rule test to assert `serializedValue` with `toBe` on stored results, matching the exact-assertion convention. The metadata test below it already asserts the full `composedFrom` with `toEqual`, so the two together cover the whole return value.

petyosi (approved):

PR #218, opened 2026-08-08
Report PR AUC as not computable for single-class datasets

ayaangazali:
## What is wrong

`PrecisionRecallEvaluator` sweeps thresholds without checking the label distribution. Its two sibling threshold evaluators do:

```ts
// ROCAUCEvaluator
if (totalPositives === 0 || totalNegatives === 0) {
  return emptyResult
}

// KolmogorovSmirnovEvaluator
if (allScores.length === 0 || positiveScores.length === 0 || negativeScores.length === 0) {
  return emptyResult
}
```

Precision-recall has the same requirement. Without negatives there are no false positives, so `precision` is 1 at every threshold by construction. Without positives the recall denominator is zero and the point-level guard (`tp + fn === 0 ? 0 : ...`) pins recall to 0, collapsing the curve onto `x = 0`.

## Evidence

Three cases sharing one class, run through all three evaluators on `817fca1`:

| dataset | PR AUC | ROC AUC | KS |
|---|---|---|---|
| all positive | **1** | NaN | NaN |
| all negative | **0** | NaN | NaN |

PR also emits a curve in both rows where ROC emits none.

Neither number measures discrimination, because there is nothing to discriminate. But 1 is indistinguishable from a perfect model and 0 from a worthless one, so a dataset that drifts to a single class reports a confident score instead of "not computable". The existing coverage does not catch this: `'threshold report evaluators return empty/NaN analyses ... when inputs are absent'` exercises the no-scores path, where all three already agree.

## What this does

Returns the same empty analysis and `NaN` scalar the other two already return, so the three agree on degenerate input. Mixed-class behaviour is untouched.

Kept `NaN` rather than a numeric sentinel because that is the convention the other two evaluators and the existing no-scores path already established, and it is what the `ScalarAnalysis` consumers already handle. Reporting the value while flagging it separately would need a schema change for a case that has no meaningful value to report.

The test covers both single-class branches and asserts all three evaluators together, so the three cannot drift apart again. Verified by removing the guard, which fails that test and no others.

---

### DIFF
--- a/.changeset/pr-auc-single-class.md
+++ b/.changeset/pr-auc-single-class.md
@@ -1,0 +1,5 @@
+---
+'logfire': patch
+---
+
+Report `PrecisionRecallEvaluator` as not computable when every case shares one class. With only positive cases it returned an AUC of 1, and with only negative cases an AUC of 0, so a dataset that carries no signal read as a perfect or a worthless model. `ROCAUCEvaluator` and `KolmogorovSmirnovEvaluator` already return `NaN` and no curve for the same input.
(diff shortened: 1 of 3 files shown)


petyosi (approved):

PR #220, opened 2026-08-08
Keep evaluating when the progress callback throws

ayaangazali:
## What is wrong

`reportProgress` calls the user-supplied callback with no guard:

```ts
function reportProgress(progress: EvaluateOptions['progress'], event: {...}): void {
  if (typeof progress === 'function') {
    progress(event)
  } else if (progress === true) {
```

It runs inside `runOne`, which runs inside `await Promise.all(expandedCases.map(runOne))`. So a throw from the callback rejects that promise, and `evaluate()` rejects with it.

Everything after the `Promise.all` is skipped: the report evaluators, `analyses`, `computeAverages`, the assertion pass rate, and `setEvalsSpanAttributes` on the experiment span. The `cases` and `failures` already collected are discarded with the rejection.

## Evidence

Two cases, a task that succeeds for both, and a progress callback that throws, on `817fca1`:

```ts
await dataset.evaluate((s) => s.toUpperCase(), {
  progress: () => { throw new Error('progress boom') },
})
```

rejects with `Error: progress boom`. Both cases had already completed successfully and their results are lost, as is the whole report.

The blast radius is the point: the callback is a reporting hook, but it takes down work that already succeeded, and the further into a long run it throws the more is thrown away.

## What this does

Catches and logs, leaving the run intact. The report evaluator loop a few lines below already isolates failures this way rather than letting one evaluator abort the experiment.

Logged via `console.error` to match the built-in `progress: true` branch in the same function. `logfire-api` does not use `diag` anywhere, so reaching for it here would introduce a reporting channel this package does not otherwise have.

Kept the failure out of the returned report. `report_evaluator_failures` is a record of evaluation results, and a progress hook produces none, so adding it there would widen a published shape for a caller-side bug.

The test asserts the run survives: both cases present, no failures, report evaluator analyses still produced, and one logged message per case. Verified by removing the guard, which fails that test and no others.

---

### DIFF
--- a/.changeset/evaluate-progress-callback.md
+++ b/.changeset/evaluate-progress-callback.md
@@ -2,4 +2,4 @@
 'logfire': patch
 ---
 
-Stop a throwing `progress` callback from discarding an entire `Dataset.evaluate()` run. The callback ran unguarded inside the `Promise.all` over cases, so one throw rejected `evaluate()` and lost the results of every case that had already finished, along with the report evaluators, analyses, averages and the experiment span's closing attributes. Progress reporting is now best effort and logs the failure instead.
+Stop a throwing `progress` callback from discarding an entire `Dataset.evaluate()` run. The callback ran unguarded inside the `Promise.all` over cases, so one throw rejected `evaluate()` and lost the results of every case that had already finished, along with the report evaluators, analyses, averages and the experiment span's closing attributes. An async callback was worse: its rejection escaped as an unhandled rejection, which terminates the process under Node's default. Progress reporting is now best effort for both, and logs the failure instead.
--- a/.changeset/evaluate-progress-callback.md
+++ b/.changeset/evaluate-progress-callback.md
@@ -1,0 +1,5 @@
+---
+'logfire': patch
+---
+
+Stop a throwing `progress` callback from discarding an entire `Dataset.evaluate()` run. The callback ran unguarded inside the `Promise.all` over cases, so one throw rejected `evaluate()` and lost the results of every case that had already finished, along with the report evaluators, analyses, averages and the experiment span's closing attributes. Progress reporting is now best effort and logs the failure instead.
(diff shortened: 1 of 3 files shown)


ayaangazali:
Good catch, and it was worse than the sync case. `progress` is declared as `(event) => void`, but an async callback is assignable to that in TypeScript, so the rejection escaped the `try`/`catch` entirely and surfaced as an unhandled rejection after the case had already completed. Under Node's default that terminates the process, where the synchronous throw only rejected `evaluate()`.

Confirmed on the branch before fixing: the run resolved with its one case intact, and `process.on('unhandledRejection')` still received `Error: async progress boom`.

Now reads the result as `unknown` and subscribes to the rejection when it is thenable, so both forms report through the same path. Added a regression that asserts no unhandled rejection fires, the case still completes, and one message is logged. Verified by removing the subscription, which brings the unhandled rejection back and fails that test and no others.

One note on the test: the repo's own lint (`no-misused-promises`, `strict-void-return`) rejects passing an async callback to a void-returning option, so the test casts to reproduce what a consumer can still do, since `tsc` allows it.

ayaangazali:
Both fixed, and the thenable one was a real hole in my own fix.

Detecting a thenable by reading `.then` invokes a getter, so a thenable with a throwing `then` getter, or a throwing `then()`, threw outside the `try` and rejected `evaluate()` again. Confirmed on the branch before changing it:

```
throwingGetter -> REJECTED then getter boom
throwingThen   -> REJECTED then boom
```

That is the same failure this PR exists to prevent. Function-valued thenables were also missed, since the guard required `typeof value === 'object'`.

Now uses `Promise.resolve(outcome)` so the assimilation happens inside promise machinery and both throws arrive at the catch handler. All three shapes now resolve with the run intact. Guarded on a non-nullish result so the ordinary void return does not allocate a promise per case.

Also replaced the fixed delay with a promise the `console.error` mock resolves, so the test waits on the logged failure rather than the scheduler.

Regressions now cover the async rejection plus both hostile thenables. Verified by reverting to the naive `.then` access, which fails exactly the two thenable cases and leaves the rest passing.

ayaangazali:
Fair, the claim was in the description but not in the fixtures. Added a function with a throwing `then`.

It is a useful case to pin, because it fails differently from the object-valued ones. Reverting to a `typeof outcome === 'object'` guard now fails three cases rather than two: the two object thenables reject `evaluate()`, while the callable one is silently ignored and simply never reports the failure. Only the second mode is invisible without this fixture.

Four hostile shapes covered now: rejecting async callback, throwing `then` getter, throwing `then()`, and callable thenable with a throwing `then()`.

petyosi (approved):

PR #221, opened 2026-08-09
Keep result map entries when the map has a value key

ayaangazali:
## What is wrong

`EvaluatorOutput` allows either a scalar, an `EvaluationReason`, or a result map:

```ts
export type EvaluatorOutput = boolean | EvaluationReason | number | Record<string, boolean | EvaluationReason | number | string> | string
export interface EvaluationReason { reason?: string; value: boolean | number | string }
```

Both of the object forms are plain objects, so `evaluationResultsFromOutput` has to tell them apart at runtime. It does so by testing for a `value` key alone:

```ts
export function isEvaluationReason(value: unknown): value is EvaluationReason {
  return typeof value === 'object' && value !== null && 'value' in value && !Array.isArray(value)
}
```

A result map whose keys include `value` matches that test, so it is read as a single `EvaluationReason` and every sibling key is discarded.

## Evidence

An evaluator returning a legal `Record<string, number>` on `817fca1`:

```ts
evaluate(): Record<string, number> {
  return { confidence: 0.9, value: 0.8 }
}
```

produces

```
scores: { MapWithValueKey: { name: 'MapWithValueKey', value: 0.8, reason: null } }
```

One score instead of two. `confidence` is gone, with no failure recorded and nothing on the span, so the case reads as fully evaluated. `value` is an ordinary name for a metric, which is what makes this reachable rather than theoretical.

## What this does

Discriminates on the key set instead of on a single key. `EvaluationReason` declares exactly `value` and optional `reason`, so an object carrying any other key is a result map:

```ts
function isSoleEvaluationReason(value: unknown): value is EvaluationReason {
  return isEvaluationReason(value) && Object.keys(value).every((key) => key === 'reason' || key === 'value')
}
```

Applied only to the top-level discrimination. The exported `isEvaluationReason` and the per-entry coercion in `buildEvaluationResultJson` are unchanged on purpose: tightening there would make an entry like `{ value: 1, extra: 2 }` fall through with the whole object in the `value` field, which the platform strict-parses, so it trades a dropped key for an invalid wire shape.

`{ value, reason }` and `{ value }` still resolve to a single result, and a map whose entries are reasons, such as `{ judge_score: { reason, value } }`, is unaffected. The test covers both branches. Verified by restoring the old check, which fails it and nothing else.

---

### DIFF
--- a/.changeset/evaluator-result-map-value-key.md
+++ b/.changeset/evaluator-result-map-value-key.md
@@ -2,4 +2,4 @@
 'logfire': patch
 ---
 
-Keep every entry of an evaluator result map that happens to contain a `value` key. A map such as `{ value: 0.8, confidence: 0.9 }` was read as a single `EvaluationReason`, so it produced one result named after the evaluator and the sibling keys were dropped without warning. A result map and an `EvaluationReason` are now told apart by their keys, since `EvaluationReason` declares only `value` and optional `reason`.
+Keep every entry of an evaluator result map that happens to contain a `value` key. A map such as `{ value: 0.8, confidence: 0.9 }` was read as a single `EvaluationReason`, so it produced one result named after the evaluator and the sibling keys were dropped without warning. A result map and an `EvaluationReason` are now told apart by shape rather than by key names alone: a lone `EvaluationReason` must carry only `value` and `reason`, with a scalar `value` and a string or absent `reason`. Key names by themselves were not enough, because `{ value: 0.8, reason: 0.9 }` is a legal map of scores.
--- a/.changeset/evaluator-result-map-value-key.md
+++ b/.changeset/evaluator-result-map-value-key.md
@@ -1,0 +1,5 @@
+---
+'logfire': patch
+---
+
+Keep every entry of an evaluator result map that happens to contain a `value` key. A map such as `{ value: 0.8, confidence: 0.9 }` was read as a single `EvaluationReason`, so it produced one result named after the evaluator and the sibling keys were dropped without warning. A result map and an `EvaluationReason` are now told apart by their keys, since `EvaluationReason` declares only `value` and optional `reason`.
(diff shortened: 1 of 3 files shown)


petyosi (changes requested):
The new discriminator still loses one legal result-map shape.

`packages/logfire-api/src/evals/evaluatorResults.ts:21`: Key names alone misclassify a valid result map whose only keys are `value` and `reason`. For example, a `Record<string, number>` containing `{ value: 0.8, reason: 0.9 }` enters the scalar branch, drops the `reason` score, renames `value` to the evaluator default, and emits numeric `reason` data even though the wire field is string or null.

Please validate the value shapes as well. At minimum, a non-string `reason` must make this a map. Add this exact collision to the regression test.

ayaangazali:
You are right, and the collision is worse than a misname. `{ value: 0.8, reason: 0.9 }` carries only the two declared keys, so it took the scalar branch: the `reason` score was dropped, `value` was renamed to the evaluator default, and a number reached the `reason` wire field, which is `string | null`.

Now checks the shapes rather than just the key names, since `EvaluationReason` declares a scalar `value` and an optional string `reason`:

```ts
return Object.entries(value).every(
  ([key, entry]) =>
    (key === 'value' && isScalar(entry)) || (key === 'reason' && (typeof entry === 'string' || entry === undefined))
)
```

That exact map is now in the regression, and it is a useful case to pin because it fails differently from the others: it produces `reason: 0.9` and `value_2: 0.8` as separate scores, so both survive and nothing non-string reaches the reason field. `{ value, reason: 'text' }` still resolves to a single result. Verified by reverting to the key-name check, which fails that test and no others.

Also rebased onto current main.

ayaangazali:
Fair, the release note described the earlier version. It now says the two are told apart by shape rather than key names alone, that a lone `EvaluationReason` needs a scalar `value` and a string or absent `reason`, and names `{ value: 0.8, reason: 0.9 }` as the case key names alone get wrong.

PR #222, opened 2026-08-09
Honour LOGFIRE_SEND_TO_LOGFIRE=false

ayaangazali:
## What is wrong

`resolveSendToLogfire` handles `'if-token-present'` and then falls through to truthiness:

```ts
const sendToLogfireConfig = option ?? env['LOGFIRE_SEND_TO_LOGFIRE'] ?? 'if-token-present'

if (sendToLogfireConfig === 'if-token-present') {
  ...
} else {
  return Boolean(sendToLogfireConfig)
}
```

An environment variable is always a string, and `Boolean('false')` is `true`. `SendToLogfire` is `'if-token-present' | boolean | undefined`, so `false` is a documented value, and `docs/reference/environment-variables.md` lists the variable as setting sending behaviour.

## Evidence

On `817fca1`, with a token present:

| input | result |
|---|---|
| `LOGFIRE_SEND_TO_LOGFIRE=false` | **true** |
| `LOGFIRE_SEND_TO_LOGFIRE=true` | true |
| `sendToLogfire: false` in code | false |

So the environment opt-out never takes effect and it fails open, continuing to send. The code path works because that `false` reaches `Boolean()` as a real boolean rather than a string, which is why the two disagree.

The existing tests only ever pass `sendToLogfire: false` through `configure()`, so the string form was never exercised.

## What this does

Reads the documented `'true'` and `'false'` strings explicitly before the truthiness fallback, trimmed and case-insensitive.

Scoped deliberately to those two. Other strings such as `'0'` and `'no'` still resolve to `true` exactly as they do today: they are not in the type, and widening the accepted set is a parity question against the Python SDK rather than a bug fix. Happy to broaden it in this PR if you would rather match Python's set.

`'if-token-present'` and the code path are untouched, and the test covers both those and the precedence of an explicit option over the environment. Verified by removing the string handling, which fails that test and no others.

---

### DIFF
--- a/.changeset/send-to-logfire-env-false.md
+++ b/.changeset/send-to-logfire-env-false.md
@@ -2,4 +2,4 @@
 'logfire': patch
 ---
 
-Honour `LOGFIRE_SEND_TO_LOGFIRE=false`. The environment variable arrives as a string and was passed to `Boolean()`, and `Boolean('false')` is `true`, so the documented way to turn sending off left it on. Setting `sendToLogfire: false` in code was unaffected. The `true` and `false` strings are now read explicitly, case-insensitively and trimmed.
+Honour `LOGFIRE_SEND_TO_LOGFIRE=false`. The environment variable arrives as a string and was passed to `Boolean()`, and `Boolean('false')` is `true`, so the documented way to turn sending off left it on. Setting `sendToLogfire: false` in code was unaffected. `true`, `false` and `if-token-present` are now normalized once and matched explicitly, case-insensitively and trimmed, so a differently cased sentinel no longer falls through to truthiness either.
--- a/.changeset/send-to-logfire-env-false.md
+++ b/.changeset/send-to-logfire-env-false.md
@@ -1,0 +1,5 @@
+---
+'logfire': patch
+---
+
+Honour `LOGFIRE_SEND_TO_LOGFIRE=false`. The environment variable arrives as a string and was passed to `Boolean()`, and `Boolean('false')` is `true`, so the documented way to turn sending off left it on. Setting `sendToLogfire: false` in code was unaffected. The `true` and `false` strings are now read explicitly, case-insensitively and trimmed.
--- a/docs/reference/environment-variables.md
+++ b/docs/reference/environment-variables.md
@@ -18,7 +18,7 @@ description: Environment variables used by the Logfire TypeScript SDK packages.
 | `LOGFIRE_ENVIRONMENT`         | Deployment environment resource metadata.                                |
 | `LOGFIRE_CONSOLE`             | Set to `true` to also print spans to the console. Boolean-only.          |
 | `LOGFIRE_MIN_LEVEL`           | Minimum manual Logfire level to emit.                                    |
-| `LOGFIRE_SEND_TO_LOGFIRE`     | Set sending behavior. `if-token-present` sends only when a token exists. |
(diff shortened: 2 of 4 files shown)


petyosi (dismissed):
The code fix is sound, but the public configuration contract needs the corresponding documentation update.\n\n: This changes the accepted syntax for  to trimmed, case-insensitive  and , but  still documents only . The repository guide requires documentation updates for public configuration changes. Please document the accepted JavaScript values and normalization so users do not assume Python forms such as , , , and  are supported here.

petyosi (commented):
Correction to my preceding review, whose code identifiers were stripped during submission:

`packages/logfire-api/src/logfireApiConfig.ts:156`: This changes the accepted syntax for `LOGFIRE_SEND_TO_LOGFIRE` to trimmed, case-insensitive `true` and `false`, but `docs/reference/environment-variables.md` still documents only `if-token-present`. The repository guide requires documentation updates for public configuration changes.

Please document the accepted JavaScript values and normalization so users do not assume Python forms such as `1`, `t`, `0`, and `f` are supported here.

petyosi (changes requested):
Correction to my preceding review, whose code identifiers were stripped during submission:

`packages/logfire-api/src/logfireApiConfig.ts:156`: This changes the accepted syntax for `LOGFIRE_SEND_TO_LOGFIRE` to trimmed, case-insensitive `true` and `false`, but `docs/reference/environment-variables.md` still documents only `if-token-present`. The repository guide requires documentation updates for public configuration changes.

Please document the accepted JavaScript values and normalization so users do not assume Python forms such as `1`, `t`, `0`, and `f` are supported here.

ayaangazali:
Documented. The table row now reads `true`, `false`, or `if-token-present`, and there is a paragraph alongside the existing `LOGFIRE_MIN_LEVEL` and `LOGFIRE_CONSOLE` notes:

> `LOGFIRE_SEND_TO_LOGFIRE` accepts `true`, `false`, or `if-token-present`, which sends only when a token is present and is the default. Values are matched case-insensitively and surrounding whitespace is ignored. The Python SDK spellings `1`, `0`, `t`, and `f` are not recognised here, and any other value enables sending.

The last clause is deliberate: an unrecognised value still falls through to truthiness, so it enables sending. That is the pre-existing behaviour and this PR does not change it, but it is the part most likely to surprise someone coming from Python, so it seemed worth stating rather than leaving implied.

Also rebased onto current main.

ayaangazali:
Both correct, and the sentinel one was mine: trimming and lowercasing only the boolean spellings left `if-token-present` on a strict comparison, so `IF-TOKEN-PRESENT` or a padded spelling missed the branch and fell through to truthiness, enabling sending with no token. That is the same failure this PR exists to fix, just moved.

Now normalizes once before matching, which also drops a level of nesting:

```ts
const configured = option ?? env['LOGFIRE_SEND_TO_LOGFIRE'] ?? 'if-token-present'
const sendToLogfireConfig = typeof configured === 'string' ? configured.trim().toLowerCase() : configured
```

Tests cover `IF-TOKEN-PRESENT` and `' if-token-present '` with and without a token. Verified by removing the normalization, which fails them and nothing else.

On the docs, you are right that `Boolean('')` is `false`, so the sentence was wrong. It now reads that an empty or blank value disables sending and any other unrecognised value enables it. Blank is included because the trim makes `'   '` behave as empty, and there is a test for that too.

PR #231, opened 2026-08-15
Remove the processor check that cannot return true

### LINKED ISSUE #228: isProcessorInstalledOnGlobal can never return true
## What is wrong

`isProcessorInstalledOnGlobal` decides whether recording is available by rejecting the proxy provider:

```ts
export function isProcessorInstalledOnGlobal(): boolean {
  const tp = TraceAPI.getTracerProvider()
  return typeof tp === 'object' && tp.constructor.name !== 'NoopTracerProvider' && tp.constructor.name !== 'ProxyTracerProvider'
}
```

`getTracerProvider()` always hands back a `ProxyTracerProvider`. From `@opentelemetry/api@1.9.1`:

```js
setGlobalTracerProvider(provider) {
  const success = registerGlobal(API_NAME, this._proxyTracerProvider, DiagAPI.instance());
  if (success) {
    this._proxyTracerProvider.setDelegate(provider);
  }
}
getTracerProvider() {
  return getGlobal(API_NAME) || this._proxyTracerProvider;
}
```

The registered global *is* the proxy, and the real provider becomes its delegate. So the second exclusion always matches and the function always returns `false`.

## Evidence

```ts
const before = isProcessorInstalledOnGlobal()                       // false
trace.setGlobalTracerProvider(new BasicTracerProvider({ spanProcessors: [getEvalsSpanProcessor()] }))
const after = isProcessorInstalledOnGlobal()                        // false
trace.getTracerProvider().constructor.name                          // 'ProxyTracerProvider' both times
```

Both call sites read it the same way:

```
Dataset.ts:477   if (capturedSpans.length === 0 && !isProcessorInstalledOnGlobal()) {
online.ts:311    capturedSpans.length === 0 && !isProcessorInstalledOnGlobal()
```

`!false` is always true, so both conditions reduce to `capturedSpans.length === 0`.

Nothing is visibly broken today, because in `Dataset.evaluate` the execute span is started inside the bucket, so a working setup captures at least one span and the length check carries the decision on its own. The problem is that the guard is silently inert, so the intent in its docstring is not actually enforced and anything later relying on it would be wrong.

## Two ways to go, and I did not want to guess

1. **Make the detection real.** Would need something other than `constructor.name`, since the proxy's delegate is private. Changes behaviour: a configured provider that captured no spans would stop reporting `SpanTreeRecordingError`, which is arguably the intended outcome for the `online.ts` path.
2. **Drop it.** Remove the dead term from both call sites and the function, leaving `capturedSpans.length === 0` as the explicit condition. No behaviour change, less code.

Happy to send either. `constructor.name` is also minification-sensitive, which is worth noting given the registry deliberately avoids relying on class names.

Built this analysis with Claude Code's help and verified the behaviour against the installed `@opentelemetry/api` myself.

ayaangazali:
Closes #228, taking option 2 from that issue. Happy to switch to option 1 instead if you would rather the detection actually work.

## What is wrong

`isProcessorInstalledOnGlobal` decides recording is available by excluding the proxy provider:

```ts
const tp = TraceAPI.getTracerProvider()
return typeof tp === 'object' && tp.constructor.name !== 'NoopTracerProvider' && tp.constructor.name !== 'ProxyTracerProvider'
```

`@opentelemetry/api@1.9.1` registers the proxy as the global and gives it the real provider as a delegate:

```js
setGlobalTracerProvider(provider) {
  const success = registerGlobal(API_NAME, this._proxyTracerProvider, DiagAPI.instance());
  if (success) { this._proxyTracerProvider.setDelegate(provider); }
}
getTracerProvider() { return getGlobal(API_NAME) || this._proxyTracerProvider; }
```

So `getTracerProvider()` returns a `ProxyTracerProvider` whether or not anything was registered, the second exclusion always matches, and the function always returns `false`. Confirmed by registering a `BasicTracerProvider` carrying the evals processor and reading it back: `false` before and after, `constructor.name` `'ProxyTracerProvider'` both times.

## What this does

Both call sites read it the same way:

```
Dataset.ts:477   if (capturedSpans.length === 0 && !isProcessorInstalledOnGlobal()) {
online.ts:311    capturedSpans.length === 0 && !isProcessorInstalledOnGlobal()
```

`!false` is always true, so both already reduced to `capturedSpans.length === 0`. This drops the term and the function, leaving that condition explicit.

No behaviour change, which is why there is no test and no changeset: the removed term could not affect either branch, and the helper is internal to `evals/spanTree`, not part of the `logfire/evals` export list. The existing suite passes unchanged.

Worth noting for option 1 if you go that way: `constructor.name` is a poor signal here regardless. The repo already treats it as unreliable, defaulting `errorFingerprinting` to false in the browser because "minified code produces unstable fingerprints", and `canonicalizeError` builds from `error.constructor.name`.

---

### DIFF
--- a/packages/logfire-api/src/evals/Dataset.ts
+++ b/packages/logfire-api/src/evals/Dataset.ts
@@ -59,7 +59,7 @@ import { Semaphore } from './Semaphore'
 import { buildDatasetJsonSchema, datasetFromObject, datasetToObject, parseYaml, stringifyYaml } from './serialization'
 import type { FromOptions, JsonSchema, ToOptions } from './serialization'
 import type { SpanTree } from './spanTree'
-import { buildSpanTree, getEvalsSpanProcessor, isProcessorInstalledOnGlobal, SpanTreeRecordingError } from './spanTree'
+import { buildSpanTree, getEvalsSpanProcessor, SpanTreeRecordingError } from './spanTree'
 
 export interface DatasetOptions<Inputs, Output, Metadata = unknown> {
   cases?: readonly Case<Inputs, Output, Metadata>[]
@@ -474,7 +474,7 @@ async function runOneCase<Inputs, Output, Metadata>(args: {
     // Build span tree and auto-extract metrics
     const capturedSpans = evalsProcessor.drainBucket(exporterContextId)
     let spanTree: SpanTree
-    if (capturedSpans.length === 0 && !isProcessorInstalledOnGlobal()) {
+    if (capturedSpans.length === 0) {
       spanTree = buildSpanTree([], new SpanTreeRecordingError())
     } else {
       spanTree = buildSpanTree(capturedSpans, null)
--- a/packages/logfire-api/src/evals/online.ts
+++ b/packages/logfire-api/src/evals/online.ts
@@ -24,7 +24,7 @@ import { emitEvaluationResult, emitEvaluatorFailure, spanReferenceFromSpan } fro
 import type { SpanReference } from './otelEmit'
 import { Semaphore } from './Semaphore'
 import type { SpanTree } from './spanTree'
-import { buildSpanTree, getEvalsSpanProcessor, isProcessorInstalledOnGlobal, SpanTreeRecordingError } from './spanTree'
+import { buildSpanTree, getEvalsSpanProcessor, SpanTreeRecordingError } from './spanTree'
 
 export type SamplingMode = 'correlated' | 'independent'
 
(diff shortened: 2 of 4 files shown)


ayaangazali:
Rebased onto `eb3a272`, green there. The diff is now 4 files rather than the whole backlog it showed against the old base.

Still reproduces: `isProcessorInstalledOnGlobal` is called at `Dataset.ts:477` and `online.ts:311` on current main, and it cannot return true at either, since the processor is installed on the evals provider rather than the global one. This deletes the predicate and the two dead `&&` terms.

Independent of #242, which touches `TailSamplingProcessor` rather than `spanTree/exporter`. Happy to close if you would rather leave the predicate in place.

petyosi:
Merged locally and landed on main as 8d2c12c — thanks! (Squash-merged outside GitHub, so closing manually.)

PR #234, opened 2026-08-16
Keep truncated Contains reasons valid UTF-8

ayaangazali:
## What is wrong

`truncatedRepr` cuts the repr on UTF-16 code units, at both ends:

```ts
return `${repr.slice(0, Math.floor(maxLength / 2))}...${repr.slice(-Math.floor(maxLength / 2))}`
```

An astral character is two code units, so a boundary falling between its halves keeps only one of them. A lone surrogate has no UTF-8 encoding, and the result goes into `reason` on the `EvaluationResult`, which is serialized into span attributes.

The callers use `maxLength` of 30, 100 and 200, so the head cut lands at index 15, 50 or 100. That is well inside the visible part of ordinary output rather than far out in a long payload.

## Evidence

`Contains` against a string output on `79bcac0`, with an emoji positioned at the head boundary:

```ts
new Contains({ value: 'ZZZNOTFOUND' }).evaluate(ctx('a'.repeat(48) + '😀' + 'b'.repeat(80)))
```

The returned `reason` contains one unpaired surrogate, and `JSON.stringify` on it emits a `\ud…` escape. Encoding it as UTF-8 and decoding back does not round-trip, because the encoder substitutes `U+FFFD`.

The tail slice has the mirror problem: its start can land on the low half of a pair, dropping the high half.

## What this does

Steps past the boundary when it falls inside a pair, at whichever end is affected, so the whole character is dropped rather than half of it.

Kept it to two code-unit checks rather than iterating code points or reaching for `Intl.Segmenter`: only the two boundaries can ever split, and the result stays within `maxLength`.

The test asserts a UTF-8 round trip rather than counting surrogates, which states the property directly and avoids spreading a string. Covers both boundaries, and verified by removing the checks, which fails it and nothing else.

---

### DIFF
--- a/.changeset/contains-repr-surrogate.md
+++ b/.changeset/contains-repr-surrogate.md
@@ -1,0 +1,5 @@
+---
+'logfire': patch
+---
+
+Stop `Contains` leaving a lone surrogate in a truncated failure reason. The reason string was cut on UTF-16 code units at both ends, so an astral character straddling either boundary lost half of itself, and the resulting reason is not valid UTF-8 once the evaluation result is serialized.
--- a/packages/logfire-api/src/evals/__test__/builtins.test.ts
+++ b/packages/logfire-api/src/evals/__test__/builtins.test.ts
@@ -181,8 +181,13 @@ describe('built-in evaluator edge cases', () => {
     const tail = `${'a'.repeat(80)}\u{1F600}${'b'.repeat(48)}`
 
     for (const value of [head, tail]) {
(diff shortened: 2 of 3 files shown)


ayaangazali:
Right, the fallback made it pass vacuously: if the containment ever stopped failing, `result.reason ?? ''` would encode an empty string and the round trip would still hold.

Now asserts `result.value` is `false` and `result.reason` is a string before encoding it, so the test cannot pass without the truncated reason it exists to check.

PR #239, opened 2026-08-17
Run the CLI when the bin is invoked through a symlink

### LINKED ISSUE #238: CLI is a silent no-op when invoked via the npm bin: npx logfire auth exits 0 and does nothing
## Summary

The CLI never runs when it is invoked through the npm bin. `npx logfire auth` exits **0 and prints nothing**, so `logfire auth`, `logfire projects use`, and the rest are unreachable by the documented commands.

Exit 0 with no output is the difficult part: there is no error to search for. A user follows the setup guide, gets no credentials, sees nothing wrong, and then finds no data in their project.

## Reproduction

```console
$ mkdir /tmp/repro && cd /tmp/repro
$ npm install logfire@0.21.8

$ ./node_modules/.bin/logfire auth ; echo "exit=$?"
exit=0                       # no output at all

$ node node_modules/logfire/dist/cli.js auth ; echo "exit=$?"

Welcome to Logfire!
Before you can send data to Logfire, we need to authenticate you.

Logfire is available in multiple data regions. Please select one:
1. US (GCP region: us-east4)
...
```

Same result via `npx logfire auth` (pinned or unpinned) and with a real TTY, so it is not a non-interactive-shell check.

| invocation | exit | output |
| --- | --- | --- |
| `./node_modules/.bin/logfire auth` | 0 | none |
| `node ./node_modules/.bin/logfire auth` | 0 | none |
| `npx -y logfire auth` | 0 | none |
| `node node_modules/logfire/dist/cli.js auth` | 0 | works |
| a **copy** of `cli.js` at another path | 0 | works |

The copy working is what rules out "the path merely differs" and points at the symlink specifically.

## Cause

`dist/cli.js` guards its entry point on:

```js
process.argv[1] !== undefined && import.meta.url === pathToFileURL(process.argv[1]).href && main()
```

npm installs the bin as a symlink (`node_modules/.bin/logfire -> ../logfire/dist/cli.js`). Node resolves `import.meta.url` to the **real** file, while `process.argv[1]` stays the **symlink** path. The two are never equal, so `main()` is never called — and because nothing throws, the process exits 0.

Minimal demonstration, independent of this package:

```js
// node_modules/.bin/guard-demo -> ../guard-demo/cli.js
import.meta.url : file:///tmp/repro/node_modules/guard-demo/cli.js
argv[1] as URL  : file:///tmp/repro/node_modules/.bin/guard-demo
guard matches   : false        // main() skipped

// invoked by real path
import.meta.url : file:///tmp/repro/node_modules/guard-demo/cli.js
argv[1] as URL  : file:///tmp/repro/node_modules/guard-demo/cli.js
guard matches   : true         // main() runs
```

## Suggested fix

Resolve the symlink before comparing:

```js
import { realpathSync } from "node:fs";

const entry = process.argv[1];
const isMain = entry !== undefined && import.meta.url === pathToFileURL(realpathSync(entry)).href;
```

Since `dist/cli.js` is a dedicated bin entry rather than a dual-purpose module, dropping the guard and always calling `main()` would also work.

## Impact beyond the package

The [TypeScript SDK get-started guide](https://pydantic.dev/docs/logfire/typescript-sdk/get-started/) prescribes exactly this sequence:

```bash
npm install @pydantic/logfire-node
npx logfire auth
npx logfire projects use my-project
```

so the documented path is currently a no-op for new users.

## Environment

- `logfire@0.21.8` (latest at time of filing)
- node v22.22.3, npm 10.9.8
- macOS (Darwin arm64)

Likely affects any platform where npm creates a symlink for the bin; Windows shims may behave differently and are untested here.

ayaangazali:
Fixes #238.

## Defect

`npx logfire auth` exits 0 and prints nothing, so every documented CLI command is unreachable through the npm bin. Nothing throws, so there is no error to search for: a user follows the setup guide, gets no credentials, and later finds no data in their project.

## Evidence

The entry-point guard at `packages/logfire-api/src/cli/index.ts:226` was:

```ts
if (process.argv[1] !== undefined && import.meta.url === pathToFileURL(process.argv[1]).href) {
```

npm installs the bin as a symlink (`node_modules/.bin/logfire -> ../logfire/dist/cli.js`). Node resolves `import.meta.url` to the real file, while `process.argv[1]` stays the symlink path, so the two are never equal and `runCli()` is never called.

Reproduced on current main (79bcac0), not only on the published 0.21.8. Symlinking a freshly built `dist/cli.js` and invoking the link prints nothing; invoking the same file by its real path prints the help.

## Fix

Resolve the symlink before comparing. The guard cannot simply be removed: `src/cli/index.test.ts` imports `runCli` from this module, so an unguarded call would run the CLI during the test suite.

The regression test symlinks the built `dist/cli.js` into a temp dir and spawns it, asserting the first line of `--help`. On the unfixed guard it fails with `expected '' to be 'The CLI for Pydantic Logfire.'`, which is the reported symptom exactly.

One tradeoff I took deliberately: `realpathSync` throws if `process.argv[1]` does not exist on disk, where the old code silently skipped `runCli()`. I did not add a try/catch, because Node refuses to start when the entry script is missing and every import path (test runner included) has a real file at `argv[1]`. Happy to wrap it if you would rather the module never throw at load.

Credit for the diagnosis goes to @strawgate in #238; I confirmed it on main and wrote the patch and test.

### DIFF
--- a/.changeset/cli-symlink-entry-guard.md
+++ b/.changeset/cli-symlink-entry-guard.md
@@ -1,0 +1,5 @@
+---
+'logfire': patch
+---
+
+Run the CLI when it is invoked through the npm bin symlink. The entry-point guard compared `import.meta.url` against `process.argv[1]`, which never match when npm installs the bin as a symlink, so `npx logfire auth` exited 0 without doing anything.
--- a/packages/logfire-api/src/cli/index.test.ts
+++ b/packages/logfire-api/src/cli/index.test.ts
@@ -1,7 +1,9 @@
 /* eslint-disable @typescript-eslint/require-await -- test stubs satisfy async signatures without awaiting. */
-import { mkdirSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from 'node:fs'
+import { spawnSync } from 'node:child_process'
+import { mkdirSync, mkdtempSync, readFileSync, rmSync, symlinkSync, writeFileSync } from 'node:fs'
 import { tmpdir } from 'node:os'
 import { join } from 'node:path'
+import { fileURLToPath } from 'node:url'
 import { Writable } from 'node:stream'
 
 import { afterEach, describe, expect, it, vi } from 'vite-plus/test'
@@ -119,6 +121,20 @@ describe('CLI entrypoint', () => {
       project_url: 'fake_project_url',
       token: 'fake_token',
     })
+  })
+
+  it('runs when the bin is invoked through a symlink', () => {
+    const dir = mkdtempSync(join(tmpdir(), 'logfire-cli-bin-'))
+    tmpDirs.push(dir)
+    const link = join(dir, 'logfire')
+    // npm installs the bin as a symlink into node_modules/.bin, which is the shape that made the
+    // entry-point guard miss and the CLI exit 0 without printing anything.
+    symlinkSync(fileURLToPath(new URL('../../dist/cli.js', import.meta.url)), link)
+
+    const result = spawnSync(process.execPath, [link, '--help'], { encoding: 'utf8' })
+
+    expect(result.status).toBe(0)
+    expect(result.stdout.split('\n')[0]).toBe('The CLI for Pydantic Logfire.')
   })
 
   function makeTmpDir(): string {
--- a/packages/logfire-api/src/cli/index.ts
+++ b/packages/logfire-api/src/cli/index.ts
@@ -1,6 +1,7 @@
 #!/usr/bin/env node
 
 import { spawn } from 'node:child_process'
+import { realpathSync } from 'node:fs'
 import { homedir } from 'node:os'
 import type { Readable, Writable } from 'node:stream'
 import { pathToFileURL } from 'node:url'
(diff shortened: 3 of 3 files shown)


strawgate:
Are you a logfire user? Did you test this against a real logfire instance?

ayaangazali:
No, I am not a Logfire user with a production instance, and I did not test against one. Worth being straight about that.

What I did test is the entry path, which is where the defect is. With the guard unfixed, invoking the built `dist/cli.js` through a symlink prints nothing and exits 0; with `realpathSync` added, the same invocation prints the help. That is `--help`, so it never touches the network. The regression test does the same thing against the built file and fails on the old guard with `expected '' to be 'The CLI for Pydantic Logfire.'`.

What I did not test is a full `logfire auth` round trip against a real backend, which needs credentials I do not have. Your report in #238 already covers that side: `node node_modules/logfire/dist/cli.js auth` reaches the region prompt, so the command works once `main()` runs. This change only decides whether `main()` runs at all, and it changes nothing inside the auth flow.

If you would rather someone with an instance verify the auth flow end to end before this merges, that seems reasonable to me.

ayaangazali:
Closing in favour of #243, which is yours and covers more. You also added the `realpathSync` try/catch I flagged as the one tradeoff I had taken deliberately, so that gap is closed too.

Thanks for picking it up.

PR #240, opened 2026-08-17
Omit the prefix from KV list query text when there is none

ayaangazali:
## Defect

Every `kv.list()` call that does not pass a prefix writes the literal text `list undefined` to `db.query.text` on the KV span. That is the default list call, so it is the common case rather than an edge case.

## Evidence

`instrumentKVFn` built the query text by interpolation:

```ts
const opts: KVNamespaceListOptions = argArray[0] || {}
const { prefix } = opts
span.setAttribute(ATTR_DB_QUERY_TEXT, `${operation} ${prefix || undefined}`)
```

The `|| undefined` idiom means "omit this" everywhere else in the file, because `setAttributes` drops undefined values. Inside a template literal it does the opposite: `undefined` stringifies, so the attribute is always present and reads `list undefined` when no prefix was given.

Verified against the file's own `createSpan` / `mockTracer` harness on current main. `instrumentKV(kv, 'CACHE').list()` set `db.query.text` to `list undefined`.

## Fix

Interpolate only when there is a prefix. The attribute stays present either way, so list spans remain consistent with every other KV operation, which always sets it. Truthiness is kept rather than an explicit `undefined` check, to match how `prefix` and the surrounding cursor and limit attributes are already treated.

The new test asserts both branches. On the old line it fails with `expected 'list undefined' to be 'list'`.

One unrelated line is included because it is required, not by choice: `kv.test.ts` was the only one of the eight test files in that directory without the `/// <reference types="@cloudflare/workers-types" />` header, so the pre-commit lint could not resolve `KVNamespace` in it and rejected the commit. Adding the header matches the other seven.

### DIFF
--- a/.changeset/kv-list-query-text.md
+++ b/.changeset/kv-list-query-text.md
@@ -1,0 +1,5 @@
+---
+'@pydantic/otel-cf-workers': patch
+---
+
+Stop writing the literal text `list undefined` to `db.query.text` on KV list spans that have no prefix.
--- a/packages/otel-cf-workers/src/instrumentation/kv.ts
+++ b/packages/otel-cf-workers/src/instrumentation/kv.ts
@@ -91,7 +91,9 @@ function instrumentKVFn(fn: Function, name: string, operation: string) {
           if (operation === 'list') {
             const opts: KVNamespaceListOptions = argArray[0] || {}
             const { prefix } = opts
-            span.setAttribute(ATTR_DB_QUERY_TEXT, `${operation} ${prefix || undefined}`)
+            // `undefined` stringifies inside a template literal, so interpolating an absent prefix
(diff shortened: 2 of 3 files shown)


ayaangazali:
Still reproduces on `4691011`: `kv.ts:94` builds the list query text as `${operation} ${prefix || undefined}`, so every `kv.list()` without a prefix, which is the default call, emits `db.query.text: "list undefined"`. The `|| undefined` idiom means "omit" a few lines above at 46-52, where `setAttributes` really does drop `undefined`, but inside a template literal it stringifies.

Green and conflict free since it was opened, and no human has looked at it yet. Happy to close it if the Cloudflare packages are not taking changes right now.

ayaangazali:

PR #241, opened 2026-08-17
Drop the stringified data point from Analytics Engine query text

ayaangazali:
## Defect

Every Analytics Engine span carries `db.query.text: "writeDataPoint [object Object]"`. `writeDataPoint` is the only method on `AnalyticsEngineDataset`, so this is on 100% of AE spans, not an edge case.

## Evidence

`instrumentAEFn` built the query text by interpolating the first argument:

```ts
span.setAttribute(ATTR_DB_QUERY_TEXT, `${operation} ${argArray[0]}`)
```

For KV and Durable Object storage that argument is a key string, so the same line reads sensibly there. For Analytics Engine it is an `AnalyticsEngineDataPoint` object, and a template literal stringifies it to `[object Object]`.

Verified on current main with the file's existing `InMemorySpanExporter` harness. `writeDataPoint({ blobs: ['b'], doubles: [1], indexes: ['idx'] })` produced exactly `writeDataPoint [object Object]`.

## Fix

Remove the attribute rather than reformat it. A data point write has no query text to report, `db.operation.name` already carries `writeDataPoint`, and the data point's contents are already summarised by `db.cf.ae.indexes`, `db.cf.ae.index`, `db.cf.ae.doubles` and `db.cf.ae.blobs`. Setting `db.query.text` to the bare operation name would only duplicate `db.operation.name`.

The new test asserts the span's complete attribute set, so it fails if the placeholder returns and also if an unexpected attribute appears. Restoring the old line fails it with `+ "db.query.text": "writeDataPoint [object Object]"`.

If you would rather AE spans keep a `db.query.text` for consistency with KV and DO storage, say so and I will set it to the operation name instead.

### DIFF
--- a/.changeset/ae-query-text-object.md
+++ b/.changeset/ae-query-text-object.md
@@ -1,0 +1,5 @@
+---
+'@pydantic/otel-cf-workers': patch
+---
+
+Stop writing `writeDataPoint [object Object]` to `db.query.text` on Analytics Engine spans. A data point write has no query text, and its contents are already summarised by the `db.cf.ae.*` attributes.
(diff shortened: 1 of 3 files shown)


ayaangazali:
Still reproduces on `118b480`: `analytics-engine.ts:47` writes `db.query.text` as `${operation} ${argArray[0]}`, and the argument is a data point object, so every Analytics Engine span carries the literal string `writeDataPoint [object Object]`. Green and conflict free since it was opened.

This is the smallest of the Cloudflare PRs I have open and it deletes a line rather than adding behaviour, so it is easy to say no to. Happy to close it if you would rather keep the attribute present, even with that value.

ayaangazali:

PR #251, opened 2026-08-19
Keep an own __proto__ key when normalizing pushed values

ayaangazali:
## Defect

`pushEvaluationDataset` silently drops an own `__proto__` key from a case value. The field is gone from the pushed dataset with no error, in a normalizer where every other awkward input raises a path-aware `DatasetSerializationError`.

## Evidence

The object branch of `normalizeValue` copied entries by assignment:

```ts
const result: Record<string, unknown> = {}
for (const [key, item] of Object.entries(value)) {
  result[key] = normalizeValue(...)
}
```

`result['__proto__'] = x` invokes the inherited `__proto__` setter rather than defining a property, so the key never lands on `result`.

An own `__proto__` key is reachable on exactly the path this function serves. An object literal cannot produce one, but `JSON.parse` does, and case values arrive from hosted exports and dataset files:

```
JSON.parse('{"a":1,"__proto__":{"nested":true}}')  ->  own keys ['a', '__proto__']
```

Reproduced on `8effa79`: pushing a case whose `inputs` is that parsed value sends `{a: 1}`.

## Fix

Build the normalized object with `Object.fromEntries`, which defines own properties. `vars/template.ts` and `vars/composition.ts` already map objects this way, so this matches the existing convention rather than introducing one, and it drops the mutable accumulator.

Reverting to the assignment loop fails the new test with `expected [ 'a' ] to deeply equal [ '__proto__', 'a' ]`.

### DIFF
--- a/.changeset/dataset-proto-key.md
+++ b/.changeset/dataset-proto-key.md
@@ -1,0 +1,5 @@
+---
+'logfire': patch
+---
+
+Keep an own `__proto__` key when normalizing values for `pushEvaluationDataset`. Assigning into a plain object invoked the `__proto__` setter instead of creating a property, so a JSON-parsed case value carrying that key had it silently dropped from the pushed dataset.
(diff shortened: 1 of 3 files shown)


petyosi:
Merged locally and landed on main as df17ff9 — thanks! (Squash-merged outside GitHub, so closing manually.)

PR #252, opened 2026-08-19
Compare baggage keys with hasOwn, not in

ayaangazali:
## Defect

A baggage entry whose key names an inherited object member is silently left off the emitted `gen_ai.evaluation.result` event. `toString`, `constructor`, `valueOf` and `hasOwnProperty` are all affected.

Baggage keys come from `propagation.getActiveBaggage()`, which is populated from the `baggage` header, so these are remote-supplied names rather than something the SDK controls.

## Evidence

`applyBaggage` skips a baggage key when the standard attributes already claim it, which is correct and matches Python's `_base_attrs` comment that "standard attributes always win on conflict". It tested for that with `in`:

```ts
for (const [k, v] of Object.entries(baggage)) {
  if (!(k in attrs)) {
    attrs[k] = v
  }
}
```

`in` walks the prototype chain, so every `Object.prototype` member reads as an existing conflict on a fresh object literal:

```
'toString' in {}        -> true
Object.hasOwn({}, 'toString') -> false
```

Reproduced on `8effa79`. With baggage `{plain, toString}`, the emitted record carried `plain` and, in place of the baggage value, the inherited `toString` function.

## Fix

Use `Object.hasOwn`. `formatter.ts` already does this deliberately in `getField`, with a comment that `Object.hasOwn` "keeps prototype members like `toString` from resolving", so this is the existing convention in the codebase rather than a new one.

Reverting to `in` fails the new test with `expected [Function toString] to be 'to-string-value'`.

Scoped deliberately: an own `__proto__` baggage key is a different problem, because it is lost at the assignment rather than the conflict check, and I could not establish that a fix there survives the logs SDK's own attribute handling. Not claiming it here.

### DIFF
--- a/.changeset/baggage-attr-keys.md
+++ b/.changeset/baggage-attr-keys.md
@@ -1,0 +1,5 @@
+---
+'logfire': patch
+---
+
+Stop dropping a baggage entry whose key names an inherited object member. `applyBaggage` tested for a conflict with `in`, so a `baggage` header carrying `toString` or `constructor` had that entry silently left off the emitted `gen_ai.evaluation.result` event.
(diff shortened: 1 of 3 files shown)


petyosi:
Merged locally and landed on main as b334f97 — thanks! (Squash-merged outside GitHub, so closing manually.)

PR #255, opened 2026-08-21
Omit the explanation attribute when a failure has no message

ayaangazali:
## Defect

An evaluator failure with no message emits `gen_ai.evaluation.explanation: ""`. pydantic-evals omits the attribute entirely, so the same failure carries an explanation on this side that is not one, and a query filtering on the attribute's presence matches it.

## Evidence

`_emit_failure` guards the assignment (`_otel_emit.py`):

```python
attrs[_ATTR_ERROR_TYPE] = failure.error_type or 'pydantic_evals.EvaluatorFailure'
if failure.error_message:
    attrs[_ATTR_EXPLANATION] = failure.error_message
```

`emitEvaluatorFailure` set it unconditionally from `failure.error_message || ''`.

The empty case is reachable rather than theoretical: `buildEvaluatorFailureRecord` uses `err.message`, and `new Error()` has an empty message. The body already handled it correctly, producing `evaluation: NAME failed` without the trailing message, so only the attribute was out of step.

Reproduced on `787e7be`: emitting a failure with `error_message: ''` produced the attribute with value `""`.

## Fix

Guard the assignment the same way. This is the same shape as #245, where a value pydantic-evals declines to record was being recorded here.

The result path is untouched and already matches: both sides check `reason` against null, so an empty-string reason is still recorded there.

This extends the existing `emits Python-compatible evaluator failure fallbacks when message/type are missing` test, which already covered this exact input but only asserted the body and error type. Reverting the guard fails it.

### DIFF
--- a/.changeset/failure-empty-explanation.md
+++ b/.changeset/failure-empty-explanation.md
@@ -1,0 +1,5 @@
+---
+'logfire': patch
+---
+
+Omit `gen_ai.evaluation.explanation` on an evaluator failure that carries no message, instead of emitting an empty string. pydantic-evals leaves the attribute off in that case, so a failure from `throw new Error()` no longer records an explanation that is not one.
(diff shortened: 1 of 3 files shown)


petyosi:
Merged locally and landed on main as 320d760 — thanks! (Squash-merged outside GitHub, so closing manually.)

PR #261, opened 2026-08-22
Respect block scope when validating template inputs

ayaangazali:
## Defect

`variablesValidate` reports a template input issue for a template that renders correctly. A bare path inside an `each` or `with` block is a property of the block's context, but validation checks it against the root `templateInputsSchema`.

```
schema:   { items: { type: 'array', items: { properties: { name: {...} } } } }
template: {{#each items}}{{name}}{{/each}}
issue:    Template path 'name' is not present in template_inputs_schema
```

Reproduced on `eb3a272`.

## Evidence

`collectPathsFromAst` walks every child node at one flat scope, so `{{name}}` in the block body is collected exactly as if it sat at the top level. `shouldValidatePath` already drops `this` and `../` paths, which are the explicit ways to write scope, but a bare path inside a context-shifting block gets no such treatment.

`referenceSyntax.ts` in the same package already models this correctly for the `@{}@` composition syntax. `collectBlockReferences` carries a `shiftedDepth`, increments it for the helpers in `CONTEXT_SHIFTING_HELPERS`, and drops a path whose own `depth` is below it:

```ts
const childShiftedDepth = CONTEXT_SHIFTING_HELPERS.has(helperName) ? shiftedDepth + 1 : shiftedDepth
collectReferencesFromAst(block['program'], references, childShiftedDepth)
collectReferencesFromAst(block['inverse'], references, shiftedDepth)
```

So the two walkers in this package disagree about block scope, and the template-input one is the one that is wrong.

## Fix

Give the template walker the same depth model, and export `CONTEXT_SHIFTING_HELPERS` so the list of helpers has one home rather than two copies that can drift.

The inverse block keeps the enclosing depth, matching Handlebars: `{{else}}` inside `{{#each}}` runs against the parent, so an unknown path there is still reported. The new test asserts that, not just the positive case.

Both halves are mutation-verified independently: removing the depth check and shifting the inverse each fail the tests. The existing `{{#if}}` case is unaffected, since `if` does not shift context.

One thing deliberately left alone: a `../` path is still dropped by `shouldValidatePath` even when the depth model could now resolve it to the root. That is a missing check rather than a false report, it predates this change, and it is a separate question.

### DIFF
--- a/.changeset/template-block-scope.md
+++ b/.changeset/template-block-scope.md
@@ -1,0 +1,5 @@
+---
+'logfire': patch
+---
+
+Stop reporting item fields inside an `each` or `with` block as missing template inputs. `variablesValidate` collected bare paths from a block body and checked them against the root `templateInputsSchema`, so `{{#each items}}{{name}}{{/each}}` was flagged for `name` even though the template renders correctly.
--- a/packages/logfire-api/src/vars/referenceSyntax.ts
+++ b/packages/logfire-api/src/vars/referenceSyntax.ts
@@ -62,7 +62,8 @@ interface Placeholder {
   value: string
 }
 
-const CONTEXT_SHIFTING_HELPERS: ReadonlySet<string> = new Set(['each', 'with'])
+/** Block helpers that run their program body against a different context. */
+export const CONTEXT_SHIFTING_HELPERS: ReadonlySet<string> = new Set(['each', 'with'])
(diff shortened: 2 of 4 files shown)


petyosi:
Merged locally and landed on main as 83cf913 — thanks! (Squash-merged outside GitHub, so closing manually.)

PR #267, opened 2026-08-24
Compare Date, RegExp, Set and Map contents in deepEqual

ayaangazali:
`deepEqual` in `evals/builtins/Equals.ts` reaches its object branch and compares own enumerable keys. A `Date`, `RegExp`, `Set` or `Map` keeps its contents in internal slots, so `Object.keys()` returns nothing for either side and any two of them compare equal.

On `118b480`:

```js
deepEqual(new Date('2020-01-01'), new Date('1990-01-01'))  // true
deepEqual(new Set([1]), new Set([2]))                      // true
deepEqual(new Map([['a', 1]]), new Map())                  // true
deepEqual(/a/, /b/)                                        // true
deepEqual(new Date(0), new Map())                          // true
```

`Equals` and `EqualsExpected` both run on `deepEqual`, and `EqualsExpected` is a default evaluator, so a task returning the wrong `Date` scores as a pass. That is the bad direction for an eval framework: the assertion goes green for a value it never looked at. Python compares these by value (`datetime(2020, 1, 1) == datetime(1990, 1, 1)` is `False`), so this is the port diverging rather than a deliberate JS choice.

The fix compares the class tag first, so two different exotic types can no longer match, then reads the contents each type actually carries: timestamp for `Date`, source and flags for `RegExp`, size and members for `Set`, size and per-key values for `Map`. Plain objects and class instances keep the existing key comparison.

`Set` members and `Map` keys are matched with `has`, which is SameValueZero. That is what a caller who put the value in the collection expects, it makes `NaN` match itself, and it avoids inventing a structural-identity rule for members that JavaScript has no hashing for. `Map` values still compare structurally.

Each branch was verified by mutation: dropping the tag check, the `Date` timestamp, the `RegExp` flags, either `size` check, the `Set` membership test, the `Map` key test, or the `Map` value test each fails one of the new assertions. 577 tests green, preflight and `pnpm run check` clean.

### DIFF
--- a/.changeset/equals-opaque-objects.md
+++ b/.changeset/equals-opaque-objects.md
@@ -1,0 +1,5 @@
+---
+'logfire': patch
+---
+
+Compare the contents of Dates, RegExps, Sets and Maps in the `Equals` and `EqualsExpected` evaluators. `deepEqual` only compared own enumerable keys, and those types keep their contents in internal slots, so any two of them matched each other and a wrong `Date` passed as equal to the expected one.
--- a/packages/logfire-api/src/evals/__test__/builtins.test.ts
+++ b/packages/logfire-api/src/evals/__test__/builtins.test.ts
@@ -54,6 +54,39 @@ describe('built-in evaluator edge cases', () => {
     expect(deepEqual({ a: 1 }, { a: 2 })).toBe(false)
   })
 
+  it('deepEqual compares the contents of Dates, RegExps, Sets and Maps', () => {
+    // A task returns these in process, so they are not constrained by the dataset file format.
+    expect(deepEqual(new Date('2020-01-01'), new Date('1990-01-01'))).toBe(false)
+    expect(deepEqual(new Date('2020-01-01'), new Date('2020-01-01'))).toBe(true)
+    expect(deepEqual(new Date(0), new Map())).toBe(false)
+    expect(deepEqual(new Date(0), { getTime: 0 })).toBe(false)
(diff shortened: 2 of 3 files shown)


ayaangazali:
Good catch, and it was worse than an incorrect result. `Symbol.toStringTag` is writable, so `Object.prototype.toString.call({[Symbol.toStringTag]: 'Date'})` returns `[object Date]`, and my branch then called `getTime()` on an object that has no such method:

```
TypeError: spoof.getTime is not a function
```

That throws out of `Equals`/`EqualsExpected` rather than returning false. The `Set` and `Map` branches had the same shape, and two spoofed `RegExp` objects compared equal because `source` and `flags` were `undefined` on both sides.

Fixed in d7ae460 by checking the prototype alongside the tag, so each branch narrows to the genuine type and anything else is simply unequal. The `as` casts are gone too, since `instanceof` narrows properly.

One tradeoff worth naming: a cross-realm `Date` (from a `vm` context or an iframe) keeps its `[object Date]` tag but fails `instanceof`, so two of those now compare unequal instead of comparing their timestamps. Being wrong in the false direction is the safe way for an assertion to fail, and it is still better than the current `main`, where a cross-realm `Date` compares equal to a `Map`.

Regressions cover all four tags, and each of the four guards was removed independently to confirm its own assertion fails.

petyosi:
Merged locally and landed on main as 3d5c543 — thanks! (Squash-merged outside GitHub, so closing manually.)

PR #268, opened 2026-08-24
Match structured attributes in span queries

ayaangazali:
`hasAttributes` compares the span attribute with `===`:

```ts
if (node.attributes[k] !== v) {
  return false
}
```

An OTel attribute value can only be a primitive or a sequence of them, so a structured value never arrives as the object the query holds. An object attribute reaches the span as the JSON string `serializeAttributes` wrote, and an array attribute arrives as a real array that is not `===` a fresh one. Both queries therefore always fail.

On `118b480`, with a span carrying `{'logfire.tags': ['a', 'b'], user: '{"id":1}'}`:

```ts
tree.any({ hasAttributes: { 'logfire.tags': ['a', 'b'] } })  // false
tree.any({ hasAttributes: { user: { id: 1 } } })             // false
tree.any({ hasAttributes: { user: '{"id":1}' } })            // true
```

Only the third works, and it makes the caller hand-write the serializer's output. pydantic-evals handles both cases in `_attribute_matches` (`otel/span_tree.py:259`): value equality first, then `json.loads(stored) == expected` when the query value is a dict or a list, with the comment that "instrumentation libraries like Logfire store dict and list values as JSON strings". This port kept the query keys but not that decoding step.

The fix routes the comparison through a helper that compares structurally, then retries against the decoded string. Decoding only runs when the query value is an object or an array, which is what keeps a string attribute `'1'` from matching the number `1`.

`deepEqual` is reused rather than reimplemented; it already backs the `Equals` evaluator in the same package.

Each branch is mutation-checked: swapping `deepEqual` back to `===` fails the array assertion, dropping the JSON retry fails the object assertion, and dropping the query-value type guard fails the string-versus-number assertion. 576 tests green, preflight and `pnpm run check` clean.

### DIFF
--- a/.changeset/span-query-json-attributes.md
+++ b/.changeset/span-query-json-attributes.md
@@ -1,0 +1,5 @@
+---
+'logfire': patch
+---
+
+Match structured attributes in span queries. `hasAttributes` compared the span attribute with `===`, so a query for an object or an array never matched: OTel carries an object as the JSON string `serializeAttributes` wrote, and an array as a real array, and neither is `===` a fresh query value. `HasMatchingSpan` and `SpanTree.any`/`find`/`first` now compare structurally and decode a JSON string attribute first, the way pydantic-evals does.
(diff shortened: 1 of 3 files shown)


petyosi:
Merged locally and landed on main as 9c6538e — thanks! (Squash-merged outside GitHub, so closing manually.)

PR #269, opened 2026-08-24
Generate a dataset schema that matches the evaluator forms

ayaangazali:
`Dataset.toText`/`toObject` write the short evaluator form, and `buildDatasetJsonSchema` does not describe it, so a dataset file this library produces fails validation against the schema this library produces.

On `118b480`:

```js
new Dataset({ cases: [new Case({ inputs: 'x', evaluators: [new Equals({ value: 1 })] })], name: 'd' }).toObject()
// cases[0].evaluators -> [{"Equals":1}]
```

The schema offers exactly two branches for `Equals`: the string `"Equals"`, and an object whose `Equals` value must be `{value, evaluation_name}`. `{Equals: 1}` matches neither, so an editor with a YAML language server marks the file invalid.

The bare-name branch is wrong in the other direction. It is emitted for every evaluator, including those with a required argument, and the loader then calls the constructor with nothing:

```js
new Equals()  // TypeError: Cannot read properties of undefined (reading 'value')
```

So the schema rejects a form that loads and accepts a form that throws.

pydantic-evals derives the branches per evaluator (`pydantic_ai/_spec.py:320-333`): the bare name only when nothing is required, `{Name: value}` when there is a single argument, and `{Name: {kwargs}}` when there is more than one. This fix does the same from each evaluator's own `jsonSchema()`, with one adjustment for the TS port: the kwargs branch is always emitted, because a TS constructor takes an options object and `decodeEvaluator` always accepts that form, and the short branch is emitted only for evaluators with a primary-argument key, since that key is what turns a bare value back into the options object.

`oneOf` became `anyOf` in the same change. `Equals` has an untyped argument, so `{Equals: {value: 1}}` now matches both its short and its long branch, and `oneOf` requires exactly one match. Python's schema is a union, which pydantic also renders as `anyOf`.

Each branch rule was mutation-checked: always emitting the bare name, dropping the short branch, giving the short branch the full argument schema, and reverting to `oneOf` each fail their own assertion. 577 tests green, preflight and `pnpm run check` clean.

### DIFF
--- a/.changeset/dataset-schema-short-form.md
+++ b/.changeset/dataset-schema-short-form.md
@@ -1,0 +1,5 @@
+---
+'logfire': patch
+---
+
+Generate a dataset JSON Schema that matches the evaluator forms the loader accepts. The schema only described `{Name: {kwargs}}`, so the short form `{Equals: 1}` that `Dataset.toText`/`toObject` writes failed validation in an editor, while a bare `Equals` validated even though constructing it with no arguments throws.
--- a/packages/logfire-api/src/evals/__test__/serialization.test.ts
+++ b/packages/logfire-api/src/evals/__test__/serialization.test.ts
@@ -22,6 +22,13 @@ import {
   stringifyYaml,
 } from '../../evals'
 
+function evaluatorBranches(schema: Record<string, unknown>, name: string): Record<string, unknown>[] {
+  const properties = schema['properties'] as { evaluators: { items: { anyOf: Record<string, unknown>[] } } }
+  return properties.evaluators.items.anyOf.filter(
+    (branch) => branch['const'] === name || Object.hasOwn((branch['properties'] ?? {}) as object, name)
+  )
+}
+
 const sleep = async (ms: number): Promise<void> =>
   new Promise((resolve) => {
     setTimeout(resolve, ms)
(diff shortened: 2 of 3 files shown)


petyosi (changes requested):
The branch derivation and the `oneOf`→`anyOf` change are right, but the stated defect class survives for `EqualsExpected`:

1. `new EqualsExpected({evaluation_name: 'x'}).toJSON()` returns a single-key object, so the encoder writes the short form `{EqualsExpected: 'x'}` — but `EqualsExpected` has no entry in the primary-arg map, so the new schema offers no short branch for it and the library again emits a file its own schema rejects. Python keys the short branch on parameter count (`_spec.py:320-338`), which covers it; keying on the hardcoded builtin map does not. Either cover `EqualsExpected` (and any single-param evaluator) or fix the encoder so it doesn't emit that form — the current encode/decode round-trip is also lossy there (`new EqualsExpected('x')` drops the name).

2. Asymmetric configuration: `buildDatasetJsonSchema`'s options accept `customEvaluators` but not `primaryArgKeys`, while decoding (`FromOptions`) accepts it — so a custom evaluator with a primary arg can be decoded but can never validate against a generated schema. Please accept `primaryArgKeys` in schema generation too.

ayaangazali:
Both addressed in `b028a77`.

**1. Single-parameter evaluators.** You were right that keying the short branch on the builtin map misses the class of defect. `encodeEvaluatorSpec` writes `{Name: value}` for *any* evaluator whose `toJSON()` returns exactly one key with a non-dict, non-list value, so the map was never the right predicate — `EqualsExpected` just happened to be the one that exposed it. It now falls back to parameter count, matching what Python keys on (`_spec.py:320-338`), so every single-parameter evaluator gets a short branch whether or not it is in the map.

The round-trip you flagged is now covered directly:

```ts
encodeEvaluatorSpec(new EqualsExpected({ evaluation_name: 'x' }))  // { EqualsExpected: 'x' }
```

and the generated schema carries the matching branch:

```json
{ "additionalProperties": false, "properties": { "EqualsExpected": { "type": "string" } },
  "required": ["EqualsExpected"], "type": "object" }
```

I left the encoder alone rather than taking your second option. `new EqualsExpected('x')` dropping the name is a constructor-overload question that is separable from schema generation, and changing the emitted form would have broken files already written by released versions. Happy to open that as its own issue if you would rather the lossy overload be fixed at the source.

**2. `primaryArgKeys` on generation.** Accepted now and merged over the builtin map, so it means the same thing on both sides of the round trip as `FromOptions`. A custom evaluator with a primary arg gets its short branch and validates against a schema generated with the same option.

Both cases went into the existing `serialization.test.ts` rather than a new file, and I checked each fails on the unfixed source rather than assuming: without the change, the `EqualsExpected` case and the `primaryArgKeys` case each fail, and the file is 32 passed with it.

PR #270, opened 2026-08-25
Reject an option value that is really the next flag

ayaangazali:
`readRequiredValue` exists five times in the CLI, and only one copy rejects a value that starts with `-`:

- `cli/commands/readTokens.ts:128` guards it, added with #249 and pinned by `index.test.ts`
- `cli/index.ts:194`, `cli/commands/whoami.ts:81`, `cli/commands/projects.ts:431` and `cli/commands/clean.ts:65` check only for `undefined`

Without the guard the next flag is consumed as the value, so the flag is silently dropped and the option is set to the flag's own name. On `4691011`:

| command | today | with this change |
| --- | --- | --- |
| `--region --base-url URL whoami` | `Unknown Logfire region "--base-url"` | `Missing value for --region` |
| `whoami --data-dir --data-dir DIR` | `Unexpected argument DIR` | `Missing value for --data-dir` |
| `projects status --data-dir --json` | `No Logfire credentials found in --json` | `Missing value for --data-dir` |
| `clean --data-dir --logs` | `No Logfire data found in --logs` | `Missing value for --data-dir` |

Every message in the middle column blames the wrong thing, and in the `status` case the `--json` a script asked for is quietly gone as well. `projects new --org --default-org NAME` is the one that reaches the network: `--org` becomes the string `--default-org` and the command goes on to list organizations, where with the guard it exits before making any request at all.

The behaviour is not in question, since the repo already chose it in #249 and has a test for it. Only four copies were left behind. This deletes them and imports one shared helper from a new `cli/args.ts`, which is a net removal of code.

The regression drives all four commands through `runCli` and asserts the exact stderr, and it fails on the unguarded helper.

### DIFF
--- a/.changeset/cli-option-value-guard.md
+++ b/.changeset/cli-option-value-guard.md
@@ -1,0 +1,5 @@
+---
+'logfire': patch
+---
+
+Reject an option value that is really the next flag in every CLI command. `read-tokens` already refused it, but the global options, `whoami`, `projects` and `clean` each had their own copy of the check without that guard, so `logfire clean --data-dir --logs` took `--logs` as the directory name and silently dropped the flag. The four copies are gone; there is one shared helper.
--- a/packages/logfire-api/src/cli/args.ts
+++ b/packages/logfire-api/src/cli/args.ts
@@ -1,0 +1,14 @@
+import { LogfireCliError } from './errors'
+
+/**
+ * Read the value that follows an option. A value starting with `-` is another option, not
+ * this one's value: consuming it would silently swallow a flag and leave the option set to
+ * the flag's own name.
+ */
+export function readRequiredValue(args: string[], index: number, option: string): string {
+  const value = args[index]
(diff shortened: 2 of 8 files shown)


ayaangazali:
This one does not hold. There is a single `const cases` in the file:

```
$ grep -n 'const cases' packages/logfire-api/src/cli/index.test.ts
317:    const cases: { args: string[]; option: string }[] = [
```

It is also scoped to that one `it` block, so even a second declaration in another test would be legal. `pnpm run check` (typecheck plus lint) passes locally and the `build` check on this PR is green, which it could not be if the file failed to compile.

Leaving the thread open rather than resolving it, since I am disputing the finding rather than acting on it. Happy to be shown otherwise if you are looking at something I am not.

petyosi:
Merged locally and landed on main as 4ced7fe — thanks! CodeRabbit's duplicate-declaration finding was checked and found incorrect. (Squash-merged outside GitHub, so closing manually.)

PR #271, opened 2026-08-25
Keep the top stack frame when there is no header line

ayaangazali:
`parseStackFrames` drops the first line of `error.stack` to skip V8's `Error: message` header. Firefox and Safari write no header, so the line it drops there is the top frame.

The file already handles those engines, `FIREFOX_PATTERN` on line 12 exists for exactly that format, but the frame that identifies the error never reaches the parser. On `4691011`, with two unrelated Firefox errors:

```js
// handleClick@http://app.example/src/ui.js:12:9      + submit@.../src/form.js:4:3
// renderChart@http://app.example/src/chart.js:88:1   + submit@.../src/form.js:4:3
canonicalizeError(first)   // "Error\n----\nsrc/form:submit"
canonicalizeError(second)  // "Error\n----\nsrc/form:submit"
computeFingerprint(first) === computeFingerprint(second)  // true
```

`computeFingerprint` is what `index.ts:818` writes to `logfire.exception.fingerprint`, so in those browsers any two errors that pass through a common caller group into one issue, and a stack of depth one fingerprints as an empty stack.

The existing Firefox test does not catch this because it mocks a stack with a V8 header line above Firefox frames, which is a shape no engine produces.

The fix keeps the first line when it parses as a frame, using the same two patterns the loop already uses rather than a new heuristic. A V8 header cannot match them: `Error: ...` fails the `at ` pattern, and the Firefox pattern needs an `@` followed by `line:column`. The one case that still skips a header is a V8 message that itself contains something like `user@host:1:2`, which would contribute one deterministic extra line. That is a far smaller cost than dropping every Firefox top frame, and grouping stays stable because the same error still produces the same string.

Regressions cover a real Firefox stack, a Safari frame with a space in the function name, and the two errors above no longer sharing a fingerprint. They fail on the unconditional `slice(1)`.

### DIFF
--- a/.changeset/firefox-top-frame.md
+++ b/.changeset/firefox-top-frame.md
@@ -1,0 +1,5 @@
+---
+'logfire': patch
+---
+
+Keep the top stack frame when fingerprinting an error in Firefox or Safari. Frame parsing dropped the first line of `error.stack` to skip V8's `Error: message` header, but those engines have no header line, so the frame that identifies the error was discarded and two unrelated errors sharing the frame below it received the same `logfire.exception.fingerprint`.
--- a/packages/logfire-api/src/fingerprint.test.ts
+++ b/packages/logfire-api/src/fingerprint.test.ts
@@ -249,6 +249,32 @@ describe('computeFingerprint', () => {
     expect(fp1).toBe(fp2)
   })
 
+  test('keeps the top frame of a stack that has no header line', () => {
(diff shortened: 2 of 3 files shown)


petyosi:
Merged locally and landed on main as c07dbea — thanks! A follow-up (8ee48bd) also matches the V8 header against the error's own name and message, closing the residual case where a frame-shaped message leaked into the fingerprint. (Squash-merged outside GitHub, so closing manually.)

PR #272, opened 2026-08-25
Stop writing an empty trace payload from the tail exporter

ayaangazali:
`TailWorkerExporter.shutdown()` calls `sendSpans([])`, which serializes a batch of zero spans and logs the result. `JsonTraceSerializer.serializeRequest([])` returns `{"resourceSpans":[]}`, so every shutdown writes an empty export into the tail stream.

That matters because of how the payload is picked up. `exportTailEventsToLogfire` forwards the first logged object that has a `resourceSpans` key, and an empty one qualifies. On `4691011`, with the empty payload logged before the real one in the same batch:

```js
const events = [{ logs: [
  { message: [{ resourceSpans: [] }] },
  { message: [{ resourceSpans: [{ scopeSpans: [{ spans: [{ name: 'real-span' }] }] }] }] },
] }]
await exportTailEventsToLogfire(events, { LOGFIRE_TOKEN: token })
// POST -> {"resourceSpans":[]}
```

The request goes out empty and `real-span` is never sent.

The fix returns early from `sendSpans` when the batch is empty, which covers `export([])` as well, and drops the `sendSpans([])` call from `shutdown` since emitting that payload was its only effect. The exporter has nothing to flush anyway: each batch is written as it arrives.

Two regressions, in a new test file next to the exporter: an empty export plus a shutdown write nothing and still report success, and a batch with a span still writes its payload. The first fails without the guard, the second fails if the guard returns unconditionally.

### DIFF
--- a/.changeset/tail-empty-export.md
+++ b/.changeset/tail-empty-export.md
@@ -1,0 +1,5 @@
+---
+'@pydantic/logfire-cf-workers': patch
+---
+
+Stop writing an empty trace payload from the tail-worker exporter. Shutdown serialized a batch of zero spans, which produced `{"resourceSpans":[]}` in the log stream, and the tail worker forwards the first logged object carrying `resourceSpans`, so an empty export could be sent in place of the batch's real spans.
--- a/packages/logfire-cf-workers/src/TailWorkerExporter.test.ts
+++ b/packages/logfire-cf-workers/src/TailWorkerExporter.test.ts
@@ -1,0 +1,63 @@
+import type { ExportResult } from '@opentelemetry/core'
+import type { ReadableSpan } from '@opentelemetry/sdk-trace-base'
+
+import { ExportResultCode } from '@opentelemetry/core'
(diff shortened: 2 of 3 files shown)


ayaangazali:
Seven days, so a status note rather than a new question.

Still reproduces on `0a592e4`. `shutdown()` calls `sendSpans([])`, which serializes to `{"resourceSpans":[]}` and logs it, and `findTraceEntry` returns the first logged object carrying a `resourceSpans` key and stops looking. So when the empty payload lands ahead of a real one in the same tail batch, the batch's actual spans are dropped.

Green, no conflicts, drift 0 on the two files it touches, and CodeRabbit approved it on the 25th.

Worth restating one thing from the body, since it is the part I would push back on if I were reviewing: restoring `sendSpans([])` in `shutdown` does not fail any test once the empty-batch guard exists, so that deletion is unpinned on purpose. It stays because emitting the payload was the call's only effect.

Happy to close this if the Cloudflare packages are parked for now. I would rather that than have it sit in your queue.

ayaangazali:

Issue #242, opened 2026-08-18
TailSamplingProcessor diverges from the Python port: distributed traces lose pending spans


petyosi:
## Summary

`TailSamplingProcessor` is a port of the Python SDK's `logfire/sampling/_tail_sampling.py`, and it diverges from it in three places. The first has real user-visible impact: **with tail sampling enabled, the Node SDK emits no pending spans at all for incoming distributed traces.**

All three are independent of #229 / #230. That pair is about buffer lifetime and, as it turns out, describes a bug that exists in the Python SDK too (filed upstream as pydantic/logfire#2273). The divergences below are ours alone and can be fixed regardless of how the upstream discussion lands.

Each one is reproduced with a test that drops straight into `packages/logfire-api/src/sampling.test.ts` and uses its existing `makeSpan` / `makeProcessor` helpers. All three pass on `79bcac0`, which is to say they confirm the current behaviour.

---

## 1. The untracked-trace path skips the deferred processor

Python has a single "no buffer" state, and it forwards to both processors (`_tail_sampling.py`):

```python
if buffer is None:
    # No buffer for this trace, meaning it was already sampled/discarded, or never tracked.
    # Deferred processor also gets on_start immediately.
    if self.deferred_processor is not None:
        self.deferred_processor.on_start(span, parent_context)
```

The port split that single state into two — `FLUSHED`, which correctly calls both processors, and `undefined`, which calls only `this.wrapped`:

```ts
// packages/logfire-api/src/TailSamplingProcessor.ts
if (!entry) {
  this.wrapped.onEnd(span)
  return
}
```

```ts
// No buffer and not root — trace started before this processor was active
this.wrapped.onStart(span, parentContext)
```

Both are missing `this.deferredProcessor?.onStart(...)` / `?.onEnd(...)`.

### Why this is not an edge case

A buffer is only created when `parentSpanContext` is absent. The server side of a distributed trace has a **remote** parent, so it never gets one, and every span in that trace takes the `undefined` path for its whole life.

In `packages/logfire-node/src/sdk.ts:367` the `PendingSpanProcessor` is reachable *only* through the tail sampler when tail sampling is on:

```ts
if (logfireConfig.sampling?.tail) {
  spanProcessors.push(
    new TailSamplingProcessor(primarySpanProcessor, logfireConfig.sampling.tail, {
      deferredProcessor: new PendingSpanProcessor(primarySpanProcessor),
    })
  )
} else {
[...]

PR #279, opened 2026-08-27
Forward forceFlush and shutdown to the deferred processor

ayaangazali:
`TailSamplingProcessor` constructs nothing but owns the deferred processor it is handed, and it never flushes or shuts it down:

```ts
async forceFlush(): Promise<void> {
  return this.wrapped.forceFlush()
}

async shutdown(): Promise<void> {
  this.buffers.clear()
  this.retainedAfterRoot.clear()
  return this.wrapped.shutdown()
}
```

Nothing else holds that reference. In `logfire-node/src/sdk.ts` the `PendingSpanProcessor` is passed straight into the constructor and is not registered with the provider, so when the SDK shuts down, the deferred processor never hears about it.

Python forwards both (`logfire/sampling/_tail_sampling.py:277-292`), calling the main processor first and then the deferred one.

Today this is a contract gap rather than a live bug: `PendingSpanProcessor.forceFlush` and `.shutdown` both just resolve. It bites the first deferred processor that has real work to finish, and the fix is two lines.

The existing "does not double-flush the wrapped processor" tests still hold, because `PendingSpanProcessor` does not delegate either call to the processor it wraps. The new test asserts that with a plain spy as the deferred processor: one call each on both sides.

Both forwards are mutation-checked, each fails the new assertion on its own. This is the third item from #242, the one with no dependency on the other two. 605 tests green, preflight and `pnpm run check` clean.

### DIFF
--- a/.changeset/tail-deferred-lifecycle.md
+++ b/.changeset/tail-deferred-lifecycle.md
@@ -1,0 +1,5 @@
+---
+'logfire': patch
+---
+
+Forward `forceFlush` and `shutdown` from `TailSamplingProcessor` to its deferred processor. The processor owns that instance, so nothing else flushes or shuts it down, and it was left running at shutdown.
--- a/packages/logfire-api/src/TailSamplingProcessor.ts
+++ b/packages/logfire-api/src/TailSamplingProcessor.ts
@@ -63,7 +63,9 @@ export class TailSamplingProcessor implements SpanProcessor {
   }
 
   async forceFlush(): Promise<void> {
-    return this.wrapped.forceFlush()
+    await this.wrapped.forceFlush()
+    // The deferred processor is owned by this one, so nothing else will flush it.
+    await this.deferredProcessor?.forceFlush()
   }
 
   onEnd(span: ReadableSpan): void {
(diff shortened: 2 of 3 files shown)


ayaangazali:
Good catch, and it matters most in the case the PR is about. Awaiting the two in sequence meant a wrapped processor that rejects skipped the deferred one entirely, so the processor this class owns went unflushed and unshut in exactly the failure it was supposed to survive.

Fixed in f7da71a. Both calls now run in order through a small helper, and the first failure is still thrown rather than swallowed, so a caller awaiting `shutdown()` sees the same rejection it saw before.

I kept the order rather than switching to `Promise.allSettled`: the deferred processor writes into the wrapped one, so starting them concurrently would change when those spans land relative to the flush they belong to. Python forwards in the same order (`_tail_sampling.py:277-292`).

Regression covers both methods on the rejection path. Three mutations, each failing it on its own: reverting either method to sequential awaits, and swallowing the failure instead of rethrowing. 606 tests green, preflight and `pnpm run check` clean.

ayaangazali:
Correction to my previous comment: it named `f7da71a`, which is the original commit on this branch, not the fix. The pre-commit hook rejected that first attempt (`throw failure` on a value typed `unknown` trips `only-throw-error`) and reverted it, and I posted the reply without checking that the commit had actually landed. The real fix is `0b6158d`, pushed now.

What it does is what I described: both calls run in order through a small helper, and the first failure is rethrown rather than swallowed. The rethrow narrows to an error object, so an error rejection passes through untouched and anything else is wrapped rather than thrown as-is.

Regression covers both methods on the rejection path. Three mutations each fail it on its own: reverting either method to sequential awaits, and dropping the rethrow. 606 tests green, preflight and `pnpm run check` clean on `6d98c9b`.

PR #280, opened 2026-08-27
Reject an unrecognised boolean environment value

ayaangazali:
`parseBooleanEnv` treats every value it does not recognise as false:

```ts
function parseBooleanEnv(value: string | undefined): boolean {
  const normalized = value?.trim().toLowerCase()
  return normalized === '1' || normalized === 't' || normalized === 'true'
}
```

So a typo picks a side silently, and for `LOGFIRE_DISTRIBUTED_TRACING` it picks the opposite of the default:

| value | today | intent |
| --- | --- | --- |
| `LOGFIRE_DISTRIBUTED_TRACING=yes` | distributed tracing off | on, and the user was trying to say so |
| `LOGFIRE_CONSOLE=on` | console off | on |
| `LOGFIRE_CONSOLE=ture` | console off | on |

Python does not guess. `_check_bool` in `config_params.py:250-260` takes `1`/`true`/`t` and `0`/`false`/`f`, and raises `LogfireConfigError` naming the value for anything else.

This does the same: the false spellings are recognised explicitly, and an unrecognised value throws with the variable name and the value. An unset or empty variable keeps its current meaning.

One behaviour change worth calling out. `LOGFIRE_CONSOLE={"enabled":true}` used to resolve to `false`, and there was a test pinning that. Object-style console config through the env var is still unsupported, so I kept the test and changed what it asserts: it now expects the error rather than a silent `false`, which read as "console deliberately off". If you would rather warn than throw on an unrecognised value, say so and I will switch it; throwing is what Python does, which is why I started there.

Each call site is mutation-checked separately, and reverting the parser to the silent-false version fails both new tests. 112 node tests green, preflight and `pnpm run check` clean.

### DIFF
--- a/.changeset/env-boolean-strict.md
+++ b/.changeset/env-boolean-strict.md
@@ -2,4 +2,4 @@
 'logfire': patch
 ---
 
-Reject a boolean environment variable whose value is neither a true nor a false spelling, instead of reading it as false. `LOGFIRE_DISTRIBUTED_TRACING=yes` silently turned distributed tracing off, and `LOGFIRE_CONSOLE=on` silently turned the console off, because anything unrecognised fell through to the false branch. Both now fail with the value in the message, matching `_check_bool` in the Python SDK, and `0`/`f`/`false` are recognised explicitly.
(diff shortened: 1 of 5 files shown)


ayaangazali:
Extended to `LOGFIRE_SEND_TO_LOGFIRE` in `logfire-api`, since it is the same rule and splitting it across two PRs would just mean reviewing the same argument twice.

That resolver ends in `return Boolean(sendToLogfireConfig)`, so an unrecognised spelling falls through to truthiness and turns sending **on**:

```js
resolveSendToLogfire({ LOGFIRE_SEND_TO_LOGFIRE: 'yes' }, undefined, token)  // true
```

So the two packages guessed in opposite directions for the same class of typo. Both now report it.

Two things kept deliberately: a boolean passed in code still short-circuits before any string handling, and a blank or whitespace-only value still means "nothing configured, do not send", which the existing test pins. Reverting the throw to `Boolean(...)` fails the new assertion.

605 api tests and 112 node tests green, preflight and `pnpm run check` clean.

PR #281, opened 2026-08-27
Count a __proto__ label or metric name in report averages

ayaangazali:
Score names, metric names and label values in a report come from evaluator output, and the averaging code assigns them into plain objects:

```ts
const dist = (labelsAcc[k] ??= {})
const lv = String(r.value)
dist[lv] = (dist[lv] ?? 0) + 1
```

`__proto__` is not a key there, it is the inherited setter, so the bucket is silently discarded. The damage is not limited to the missing entry: the remaining values are normalised against a total that no longer includes it.

On `6d98c9b`, two cases labelled `__proto__` and `good`:

```js
computeAverages('labels', [makeCase('__proto__'), makeCase('good')]).labels
// { tier: { pro: 1 } }   <- reported as 100% good, not 50%
```

`averageFromAggregates` reads those distributions back for multi-run reports, so the same wrong number rolls up. A report parsed from JSON can also carry an own `__proto__` score, metric or label name, which the accumulators drop the same way.

The fix accumulates in `Map`s and turns them into records with `Object.fromEntries`, which is what `vars/template.ts` and `datasets/json.ts` already do for the same reason. No behaviour changes for any other key.

Three regressions, one per converted site, each mutation-checked: skipping the value in the distribution, rebuilding the records by assignment, and skipping it in the roll-up each fail their own assertion. The expected distribution in the test is itself built with `Object.fromEntries`, since a `__proto__` key in an object literal would set the prototype rather than the key. 606 tests green, preflight and `pnpm run check` clean.

### DIFF
--- a/.changeset/report-labels-proto.md
+++ b/.changeset/report-labels-proto.md
@@ -1,0 +1,5 @@
+---
+'logfire': patch
+---
+
+Count a score, metric or label named `__proto__`, and a label value of `__proto__`, in report averages. Those names come from evaluator output and were assigned into plain objects, so the inherited setter swallowed them: the bucket disappeared and the remaining label values were renormalised as if it had never existed, reporting 100% for a label that held half the cases.
--- a/packages/logfire-api/src/evals/__test__/reporting.test.ts
+++ b/packages/logfire-api/src/evals/__test__/reporting.test.ts
@@ -5,6 +5,7 @@ import {
   Case,
   CaseLifecycle,
   computeAssertionPassRate,
+  averageFromAggregates,
   computeAverages,
   ConfusionMatrixEvaluator,
   Dataset,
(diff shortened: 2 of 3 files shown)


ayaangazali:
Adding a consequence found while reviewing a related change: this is not only a lost bucket, it is prototype pollution on `main`.

`computeAverages` does `scoresAcc[k] ??= { count: 0, sum: 0 }`. For `k = '__proto__'` the read finds `Object.prototype`, `??=` sees a non-nullish value and skips the assignment, and the `count`/`sum` increments then land on `Object.prototype`:

```js
acc === Object.prototype   // true
'count' in {}              // true
```

The `Map` accumulators in this PR close that as well as the miscounted distribution, and the same input on this branch leaves `Object.prototype` untouched.

It is only reachable today through a report parsed from JSON, since a result named `__proto__` currently gets renamed before it reaches a case. That changes if #284 lands, so this one wants to go first.

ayaangazali:
Landed as `2226a59` and released in #285, so closing this.

Thanks for reshaping it into `KeyedAccumulator` rather than taking the three patched assignments. Your version also fixes something mine did not: `averageFromAggregates` walked a union of label keys and read `a.labels[key]` back out, so a `__proto__` label found `Object.prototype`, counted as present in every aggregate, and inflated the divisor. Iterating `Object.entries` per aggregate removes that as a side effect.

PR #283, opened 2026-08-28
Record an absolute schema path relative to the dataset file

ayaangazali:
`Dataset.toFile` writes the schema file wherever the caller asked, but records the option verbatim as the reference inside the dataset. With an absolute `schemaPath`, that reference is an absolute path from the writer's machine.

On `6d98c9b`, writing into a temp directory:

```
# yaml-language-server: $schema=/var/folders/vy/.../T/ds-r6d6D0/cases_schema.json
"$schema": "/var/folders/vy/.../T/ds-r6d6D0/cases_schema.json"
```

The sidecar itself lands correctly, so nothing fails at write time. The cost appears later: commit that dataset, open it anywhere else, and the language server resolves the schema to a directory that does not exist, silently losing validation and completion for the file the schema was generated for.

pydantic-evals hit the same thing and fixed it in #6143: when the schema path is absolute and sits under the dataset file's own directory, the recorded reference is made relative to that directory, and anything outside it is kept as given.

This does the same. The sidecar is still written to the absolute location; only the reference changes, and only when a relative form exists:

```
# yaml-language-server: $schema=cases_schema.json
"$schema": "cases_schema.json"
```

A relative `schemaPath` and a URL are untouched, and the existing idempotent-sidecar test still passes unchanged.

Two mutations, each failing the new assertion: recording the raw option again, and stripping to the basename unconditionally, which would rewrite an unrelated absolute path into a sibling that is not there. 605 tests green, preflight and `pnpm run check` clean.

### DIFF
--- a/.changeset/dataset-absolute-schema-path.md
+++ b/.changeset/dataset-absolute-schema-path.md
@@ -1,0 +1,5 @@
+---
+'logfire': patch
+---
+
+Record an absolute `schemaPath` relative to the dataset file in `Dataset.toFile`. The absolute path was written verbatim into `$schema` and into the `yaml-language-server` comment, so a dataset committed from one machine pointed at a directory that exists nowhere else and the schema silently stopped resolving. A path outside the dataset's own directory is still written as given.
--- a/packages/logfire-api/src/evals/Dataset.ts
+++ b/packages/logfire-api/src/evals/Dataset.ts
@@ -669,12 +669,13 @@ function resolveSiblingPath(filePath: string, siblingPath: string): string {
 }
 
 /**
- * The `$schema` value to record inside the dataset. An absolute path under the dataset's own
- * directory becomes relative to it, so a committed dataset does not carry the writer's machine
- * layout; anything else is kept as given. Mirrors `Dataset.to_file` in pydantic-evals.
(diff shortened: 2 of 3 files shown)


ayaangazali:
Correct, and it broke the exact case this PR is about. `C:` matches the URI-scheme pattern I reused from `resolveSiblingPath`, so on Windows every absolute schema path was classed as a URI and skipped the relative-reference path entirely.

Fixed in 1800b7d. Once a drive letter is no longer treated as a scheme, the URI branch turns out to be unreachable: `isRooted` already covers URLs, and a URL never starts with the dataset directory's prefix, so it falls through unchanged anyway. The branch is gone rather than corrected, which is one condition fewer to reason about.

The regression runs anywhere rather than only on Windows. It drives `toFile` through the Deno stub the file already uses for filesystem tests, so the paths stay strings: a dataset at `C:\data\dataset.yaml` with `schemaPath: C:\data\dataset.schema.json` now records `dataset.schema.json`, and an `https://` schema path is still recorded verbatim.

Mutation-checked: putting the single-character scheme back fails the drive-letter assertion. 606 tests green, preflight and `pnpm run check` clean, and check re-run after the commit since the hook rewrites staged files.

PR #284, opened 2026-08-28
Keep an evaluator result named like an Object.prototype member

ayaangazali:
Evaluator result names are user-supplied, and `runEvaluators` treats the result bucket as if it had no prototype:

```ts
function nextResultName(existing: Record<string, EvaluationResultJson>, baseName: string): string {
  if (existing[baseName] === undefined) {
    return baseName
  }
  ...
}
```

`existing['toString']` finds the inherited method, so the name reads as already taken. On `6d98c9b`, an evaluator returning `{ __proto__: 0.9, ok: 1 }` as a result map:

```
case scores keys: [ '__proto___2', 'ok' ]
```

The score is reported under a name no evaluator declared, and the same happens to `toString`, `constructor`, `valueOf` and the rest. Fixing only the lookup is not enough: the write is `out.scores[name] = ...`, so a genuine `__proto__` name would then run the inherited setter and the result would vanish instead.

So this changes both halves. The collision check uses `Object.hasOwn`, matching the reasoning in #252, and the write uses `defineProperty` so the entry is stored as an own property. The three near-identical branches collapse into one while I am there, since only the bucket differed.

The regression uses `Object.fromEntries` to build the result map, which is how a real one ends up with an own `__proto__` key, an object literal would set the prototype instead. Both halves are mutation-checked: restoring the `undefined` lookup and restoring the plain assignment each fail it.

One note on the test, since it looks odd: the scores are read through a helper rather than as `scores.toString`, because the linter rewrites the index form to dot notation and that resolves to the inherited method rather than the recorded score. 605 tests green, preflight and `pnpm run check` clean.

### DIFF
--- a/.changeset/evaluator-result-proto-name.md
+++ b/.changeset/evaluator-result-proto-name.md
@@ -1,0 +1,5 @@
+---
+'logfire': patch
+---
+
+Keep an evaluator result whose name collides with `Object.prototype`. The duplicate-name check read the bucket with `existing[name]`, which finds the inherited member, so a result named `toString` or `__proto__` looked like a repeat and was renamed to `toString_2`, and writing a genuine `__proto__` name ran the inherited setter instead of storing the result. The check now uses `Object.hasOwn` and the write defines an own property.
--- a/packages/logfire-api/src/evals/__test__/offline.test.ts
+++ b/packages/logfire-api/src/evals/__test__/offline.test.ts
@@ -191,6 +191,39 @@ describe('offline evals — span attribute parity', () => {
     expect(result.cases[0]?.assertions['duplicate_2']?.value).toBe(false)
   })
 
+  it('keeps a result name that collides with Object.prototype', async () => {
+    class PrototypeNamedEvaluator extends Evaluator {
+      static override evaluatorName = 'PrototypeNamedEvaluator'
+      evaluate(): Record<string, number> {
+        // `fromEntries` is how a real result map ends up with an own `__proto__` key, the same
+        // way `JSON.parse` does; an object literal would set the prototype instead.
+        return Object.fromEntries([
+          ['__proto__', 0.9],
+          ['toString', 0.5],
+        ]) as Record<string, number>
+      }
+    }
+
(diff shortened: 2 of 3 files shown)


ayaangazali:
Confirmed, and thank you for chasing it past the diff. Reproduced on `6d98c9b`:

```js
const scoresAcc = {}
const acc = (scoresAcc['__proto__'] ??= { count: 0, sum: 0 })
acc.count += 1
acc === Object.prototype        // true
'count' in {}                   // true
```

`??=` reads the inherited value first, finds `Object.prototype`, decides it is not nullish and skips the assignment, so the `+=` lands on `Object.prototype` itself. That is real pollution, not just a lost bucket, and this PR is what would let such a name reach it.

It is already fixed in #281, which converts those accumulators in `reporting.ts` to `Map`s with `Object.fromEntries` at the end. Same input on that branch:

```
averages scores keys: [ '__proto__' ]
Object.prototype polluted? false
```

So the two need to land together, and #281 first if they land separately. I would rather not duplicate the aggregation fix here, since it would then differ from #281 during review. @petyosi, say the word if you would prefer one PR instead and I will fold #281 into this one and close it.

Leaving the thread open until that is settled rather than resolving it, since it is a real constraint on merge order rather than something this diff can answer on its own.

ayaangazali:
The ordering question here is settled: `2226a59` landed the report-averages fix, so the aggregation is prototype-safe and this one is unblocked. It is `clean` against `3d33c49` and needs no rebase.

Two notes after reading how you reshaped it.

You pulled the accumulate-then-finalize shape into `KeyedAccumulator` so prototype safety is a property of the abstraction rather than a per-site convention. This diff has three buckets, `assertions`, `scores` and `labels`, all doing the same name-then-store, and it currently guards each store individually. Happy to give them one small writer the same way if you would rather it read like `reporting.ts` does now. Say the word and I will push it rather than have you patch it after merge.

Separately, `KeyedAccumulator` is unexported in `reporting.ts`, which is the right place for it today. If a second module ever needs it, this is the one I would expect to reach for it.

PR #286, opened 2026-08-29
Restore the data directory gitignore after logfire clean

ayaangazali:
### Defect

`logfire clean` empties the data directory but leaves the directory itself in place. `ensureDataDir` only seeded the `.gitignore` when it created the directory, so the next `logfire auth` / project setup wrote `logfire_credentials.json` back into a directory with no ignore rule covering it. The data directory usually lives inside the user's repository, so the credentials file is then an untracked, unignored file that `git add .` will pick up.

### Evidence

Measured on `3d33c49`, driving the real functions:

```
after first write     [".gitignore", "logfire_credentials.json"]
after logfire clean   []                                          dir still exists: true
after second write    ["logfire_credentials.json"]
```

`removeProjectCredentials` deletes the `.gitignore` along with the two credential files (`credentials.ts:293`), which is what the `clean` confirmation prompt lists, so the deletion itself is intended. What was missing is the other half: nothing ever writes the rule back.

pydantic/logfire fixed the same thing in `ensure_data_dir_exists` two weeks ago, in #2173.

### Fix

Seed the rule for a directory that holds nothing but the files Logfire itself writes, which covers both a newly created directory and one `clean` has emptied. A directory with any other contents is left alone, because `--data-dir` can point at a directory of the user's own and `*` would ignore all of it.

The write uses exclusive creation, so an existing `.gitignore` keeps its rules and a symlink planted in a data directory that arrived from elsewhere is refused rather than written through. That matches the `lstatSync` guard already above it and the `O_NOFOLLOW` handling on the read-token path.

Two regressions: the clean-then-rewrite cycle, and a data directory holding a file of the user's own staying unignored. Three mutations, one per guard: restoring the seed-on-create-only shape, dropping the only-our-files check, and swapping exclusive creation for a plain write each fail a different assertion, the last one being the existing `does not clobber an existing .gitignore` test.

### DIFF
--- a/.changeset/gitignore-after-clean.md
+++ b/.changeset/gitignore-after-clean.md
@@ -1,0 +1,5 @@
+---
+'logfire': patch
+---
+
+Restore the data directory's `.gitignore` when credentials are written after `logfire clean`. The rule was only seeded when the directory was created, so re-authenticating into a directory that `clean` had emptied wrote `logfire_credentials.json` with no ignore rule covering it. A directory holding anything besides the files Logfire writes itself is still left alone.
--- a/packages/logfire-api/src/cli/credentials.test.ts
+++ b/packages/logfire-api/src/cli/credentials.test.ts
@@ -106,6 +106,39 @@ token = "ignored"
     expect(readFileSync(join(dir, '.gitignore'), 'utf8')).toBe('node_modules\n')
   })
 
+  it('restores .gitignore when credentials are written after logfire clean', () => {
+    const dir = join(makeTmpDir(), '.logfire')
+    const credentials = {
+      logfire_api_url: 'https://logfire-us.pydantic.dev',
+      project_name: 'myproject',
+      project_url: 'https://logfire.pydantic.dev/org/myproject',
+      token: 'write-token',
+    }
+    writeProjectCredentials(dir, credentials)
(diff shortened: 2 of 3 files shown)


ayaangazali:
Good catch, and it reproduces. Fixed in `429f3fe`.

`startsWith(READ_TOKEN_FILENAME)` accepted names Logfire never writes. Measured on the previous head:

```
read_token.jsonl      gitignore written=true
read_token.json.bak   gitignore written=true
notes.txt             gitignore written=false
```

So a user's own `read_token.jsonl` sitting alone in the directory got a `*` rule written next to it and disappeared from Git, which is the exact outcome the guard exists to prevent.

`isDataDirFile` now matches the four names this package actually creates: `.gitignore`, `logfire_credentials.json`, `read_token.json`, `read_token.json.lock`, plus `read_token.json.<uuid>.tmp` with the UUID checked against `randomUUID()`'s shape, since that is what `reserveReadTokenSave` renames from. `read_token.json..tmp` no longer counts.

Two regressions: the three lookalike names stay unignored, and a directory holding an in-flight save's lock and temporary file still gets the rule. Two mutations: restoring the prefix match, and dropping the lock and temporary names, each fail one of them.

On the ast-grep note about `execFileSync`, that import predates this PR and the diff does not touch it, so I have left it alone.

PR #294, opened 2026-08-31
Report both shutdown failures

ayaangazali:
### Defect

`shutdown()` starts the SDK teardown and the variables teardown, then awaits them with `Promise.all`. `all` rejects the moment the first one fails, so:

- the second failure is discarded, in a function that otherwise goes to some length to collect failures and raise them as an `AggregateError`, and
- the operation that had not settled yet is left running after the call returned.

The second part matters because the `finally` clears `activeRuntime` on the way out, so a following `start()` can register a new runtime while the previous variables teardown is still in flight.

### Evidence

Measured on `3d33c49` with the SDK teardown rejecting immediately and the variables teardown rejecting a tick later:

```
before: shutdown() rejects with the SDK error alone; the variables teardown is still pending
after:  AggregateError('logfire SDK: shutdown failed') carrying [sdkError, variablesError]
```

`Promise.all` rejecting early is also what the `shutdownPromise.catch(() => undefined)` line existed to silence, so the dropped rejection was known about; it was the reporting that was missing.

### Fix

`Promise.allSettled`, then push every rejection into the `errors` array the function already builds. Both operations were started eagerly before, so this waits only for work already in flight, and `withDeadline` still bounds the wait exactly as it did.

Same reasoning as #279, which you merged: when a call owns two teardowns, both should run and the first failure should not hide the second.

Two mutations: restoring `Promise.all`, or dropping the loop that collects the rejections, each fail the new assertion.

The test needed one harness addition, `queueShutdownVariablesPromise`, mirroring the existing `queueShutdownPromise`, so the variables teardown can be made to settle after the SDK one.

### DIFF
--- a/.changeset/shutdown-aggregate-both.md
+++ b/.changeset/shutdown-aggregate-both.md
@@ -2,4 +2,4 @@
 '@pydantic/logfire-node': patch
 ---
 
-Report both shutdown failures instead of the first. `shutdown()` awaited the SDK and the variables teardown with `Promise.all`, which rejects on the first failure, so the second error was dropped and the operation it belonged to was left running past the call. They are now awaited together and every failure is collected, which is what the surrounding `AggregateError` handling was already built for.
+Report both shutdown failures instead of the first. `shutdown()` awaited the SDK and the variables teardown with `Promise.all`, which rejects on the first failure, so the second error was dropped and the operation it belonged to was left running past the call. Each failure is now recorded as its own operation settles, so an early one survives even when the other teardown outlives the deadline.
--- a/.changeset/shutdown-aggregate-both.md
+++ b/.changeset/shutdown-aggregate-both.md
@@ -1,0 +1,5 @@
+---
+'@pydantic/logfire-node': patch
+---
+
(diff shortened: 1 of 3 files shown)


ayaangazali:
Both right, and the second one was a regression I introduced. Fixed in `29a93fa`.

**The deadline case.** You are correct that waiting on a combined result throws away an early failure. Measured on the previous head, with the SDK teardown rejecting immediately and the variables teardown never settling:

```
kind=Error  message=logfire SDK: shutdown timed out  errors=null
```

The SDK error is gone, and the old `Promise.all` version did report it there, so my change made that path worse rather than better. Each failure is now recorded as its own operation settles, and the deadline error is appended after them, so the reported `AggregateError` carries `['sdk shutdown failed', 'logfire SDK: shutdown timed out']`. That case is now its own test.

**The test timing.** Also right, and it made the pending-ness assertion vacuous. It now waits for `shutdownVariablesCalls` to reach 1 and asserts the promise is still unsettled before rejecting it, so the sequence is actually pinned rather than assumed.

One further change while rewriting it: the error order is no longer asserted. Errors are recorded in settle order now, which is not something worth pinning, so that assertion sorts. The deadline test does assert order, because there the deadline error is deliberately last.

Three mutations, each failing something different: restoring `Promise.all` fails both tests, and the intermediate `allSettled`-and-read-the-result version you flagged fails the new deadline test alone.

PR #297, opened 2026-09-02
Look up a variable by own name only

ayaangazali:
### Defect

`getVariableConfig` reads the variables record with a plain index, and the name comes from `defineVar`. A variable named after an `Object.prototype` member therefore resolves to the inherited member, which is returned as if it were a `VariableConfig` and then fails somewhere downstream.

### Evidence

Measured on `3ed526d`, against a config holding only `main`:

```
absentname     reason=code_default  value="fallback"
constructor    reason=other_error   value="fallback"
valueOf        reason=other_error   value="fallback"
```

The value is right either way, so nothing is corrupted, but a name nobody configured is reported as an internal failure instead of as unconfigured, and a real error is being swallowed to produce it. `other_error` is part of the resolution reason a caller can read.

This is the shape `e5e0d93` just consolidated. That pass covered the evaluator registry, result buckets and task-run state; the variables lookup is the same read in a file it did not touch.

### Fix

`getOwn`, the helper from that commit, in the one place that needed it. Reverting the read fails the new assertion.

Two notes on scope.

The alias scan below it is already safe, since it iterates `Object.values`.

I also changed `followRef`'s `config.labels[labeled.ref]` and then reverted it: mutating that line broke no test, and the reason turned out to be that `validateLabelRefs` already rejects a ref that is not an own key of `labels`, using `Object.hasOwn`. So that read cannot see an inherited member, and guarding it would have been a change with no reachable case behind it.

### DIFF
--- a/.changeset/vars-lookup-own-keys.md
+++ b/.changeset/vars-lookup-own-keys.md
@@ -1,0 +1,5 @@
+---
+'logfire': patch
+---
+
+Treat a variable named like an `Object.prototype` member as unconfigured. `getVariableConfig` read the variables record with a plain index, so `defineVar('constructor')` or `defineVar('valueOf')` found the inherited member, handed it back as if it were a variable config, and failed downstream: the value still fell back to the code default but the resolution reason was reported as `other_error` rather than `code_default`.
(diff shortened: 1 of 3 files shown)


petyosi:
Thanks — the lookup bug is real and is fixed on main in 92f0868 ("Close the prototype-key hazard class across the variables and attribute paths"), with co-author credit to you. Rather than merging the single-site fix, we swept the whole class: this PR and #298 covered two of roughly ten sites where user-controlled keys meet plain records (the variable store's updateVariable/batchUpdate, followRef, rollout normalization, the attribute scrubber, and serializeAttributes had the same defect). Your test design is carried over into the sweep.

PR #298, opened 2026-09-02
Keep a prototype-named variable or label when normalizing

ayaangazali:
### Defect

`normalizeVariablesConfig` and `normalizeLabels` build their records by plain assignment, so an entry named `__proto__` runs the inherited setter and disappears. Both run on every config the store reads back out, since `getAllVariablesConfig` goes through `cloneVariablesConfig`.

`__proto__` satisfies the SDK's own name rule, `/^[A-Za-z_][A-Za-z0-9_]*$/`, which is the pattern Python's `VariableName` uses, so it is a legal variable name rather than a hostile one.

### Evidence

Measured on `3ed526d`. Creating two variables in a `LocalVariableProvider`:

```
getAllVariablesConfig keys = ["normal"]
own __proto__ present      = false
re-create __proto__        = VariableAlreadyExistsError
```

So the store considers the variable created, and the config read back out of it does not contain it. `createVariable` itself is fine, since it uses a computed key in an object literal, which defines an own property; the loss happens on normalization.

The label case surfaces differently, as a misleading error rather than a silent gap. A config with a `__proto__` label fails to load with:

```
Label '__proto__' present in rollout.labels is not present in labels
```

for a label the payload does contain.

### Fix

Accumulate both in a `Map` and materialize with `Object.fromEntries`, which defines own keys. That is the rule `e5e0d93` set for records whose keys the SDK does not choose, applied to the two in this file.

Two mutations, one per record: reverting either normalization to plain assignment fails its own test.

Related but separate, so it is in #297 rather than here: `getVariableConfig` reads that same record with a plain index.

### DIFF
--- a/.changeset/vars-normalize-own-keys.md
+++ b/.changeset/vars-normalize-own-keys.md
@@ -1,0 +1,5 @@
+---
+'logfire': patch
+---
+
+Keep a variable or label whose name is an `Object.prototype` member when normalizing the variables config. Both records were built by plain assignment, so a `__proto__` entry ran the inherited setter and vanished: a variable the store had accepted was missing from every config read back out of it, and a dropped label made the rollout validator report it as "not present in labels" for a label the payload does contain. Such a name satisfies the SDK's own identifier rule, the same one Python uses.
--- a/packages/logfire-api/src/vars.test.ts
+++ b/packages/logfire-api/src/vars.test.ts
@@ -685,6 +685,47 @@ describe('managed variables', () => {
     expect(calls[0]?.headers.get('Content-Type')).toBeNull()
   })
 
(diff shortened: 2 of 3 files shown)


petyosi:
Thanks — fixed on main in 92f0868 with co-author credit to you, as part of a sweep of the whole prototype-key class. Your normalizeVariablesConfig/normalizeLabels fixes were correct, but normalizeRollout 60 lines below had the same bug (a __proto__ rollout weight was dropped after passing the very validation this PR unblocks), as did the store's updateVariable/batchUpdate and the label resolution reads. All sites now follow the repo convention from e5e0d93: Maps inside, Object.fromEntries at the boundary, getOwn/setOwn for single keys.

PR #299, opened 2026-09-02
Report an unregister that throws undefined

ayaangazali:
### Defect

The instrumentation-unregister loop in `startBrowserInstrumentations` used the thrown value itself as its failure sentinel:

```ts
let firstError: unknown
...
catch (error) { firstError ??= error }
...
if (firstError !== undefined) { throw ... }
```

An unregister callback is user supplied, so `undefined` and `null` are both legal things for it to throw. `undefined` leaves the accumulator nullish and reads as no failure at all, and a first `null` is overwritten by a later error through `??=`, despite the name.

This is the same defect `ecc431b` fixed in `runBoth` this morning, in a different package.

### Evidence

Measured on `3ed526d`, with the unregister step throwing `undefined`:

```
unregister threw undefined -> RESOLVED (the unregister failure was swallowed)
```

Cleanup reports success and the caller is never told an instrumentation failed to unregister. With a real `Error` it rejects, which the existing memoization test already covers, so the two paths disagree only for the values a raw sentinel cannot represent.

### Fix

A separate `failed` flag decides whether to rethrow, the same shape as `runBoth`. The two other error accumulators in this file, `rollbackError` and `firstCleanupError`, already avoid the problem by normalizing to an `Error` at capture; this was the odd one out.

One thing left unpinned, deliberately. The regression covers the `undefined` case. The `null`-overwrite half is not covered, because the test harness exposes a single failing unregister step and I could not express two failing callbacks through it. The `if (!failed)` guard stays because it is what makes the variable's name true, and mutating it away accordingly fails no test.

### DIFF
--- a/.changeset/browser-unregister-sentinel.md
+++ b/.changeset/browser-unregister-sentinel.md
@@ -1,0 +1,5 @@
+---
+'@pydantic/logfire-browser': patch
+---
+
+Report an instrumentation `unregister` that throws `undefined` instead of resolving cleanup as if it had succeeded. The accumulator held the raw thrown value and treated it as the failure sentinel, so `undefined` read as no failure at all, and a first `null` was overwritten by a later error through `??=`. Failures are now normalized to an `Error` when captured, the way the other two error accumulators in that file already do.
--- a/packages/logfire-browser/src/index.test.ts
+++ b/packages/logfire-browser/src/index.test.ts
@@ -1767,6 +1767,25 @@ describe('browser cleanup', () => {
     expect(tracerProvider.shutdownCalls).toBe(1)
   })
 
+  it('reports an unregister that throws undefined instead of resolving', async () => {
+    // An unregister callback is user supplied, so `undefined` is a legal thing for it to throw.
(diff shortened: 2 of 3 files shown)


ayaangazali:
Heads up that #303 touches the same two files, so ordering matters more than the size of this diff suggests.

The overlap looks incidental rather than semantic: #303 adds cleanup steps around `runCleanupStep('instrumentation unregister', ...)`, while this changes the accumulator inside `startBrowserInstrumentations` that the unregister step runs. Different regions of `index.ts`, so I expect them to merge, but the context lines are close enough that I would not assume it.

Happy either way: land #303 first and I will rebase this and re-run the gates, or take this first since it is four lines and a test. Your call, and no reply needed if you would rather just merge in whatever order suits.

petyosi:
Thanks — the bug is real and is fixed on main in 05ac16b with co-author credit to you, and your test carries over nearly verbatim. One change from this PR's approach: instead of the flag-plus-deferred-normalization pattern, the fix normalizes to an Error at capture, which is the idiom the other two accumulators in this file already use and keeps ??= sound with no extra state (it is also what this PR's changeset text described).

PR #304, opened 2026-09-03
Send a nested non-finite number as a string

ayaangazali:
### Defect

#292 fixed this for a top-level numeric attribute. A non-finite number nested inside an object or array attribute is still lost: `stringifyJsonAttribute` calls `JSON.stringify`, which has no spelling for `NaN` or `Infinity` and writes `null`. The generated schema then describes it as `null` too, so the two agree with each other and both disagree with what the value was.

### Evidence

Measured on `3ed526d`:

```
serializeAttributes({ nested: { inf: Infinity, nan: NaN, ok: 1.5 }, top: NaN })

nested = "{\"inf\":null,\"nan\":null,\"ok\":1.5}"
top    = "NaN"
```

The top-level value survives and the nested ones do not, in the same call. A consumer cannot tell a dropped `NaN` from an attribute that really was `null`.

Python does not have the split: `to_json_value` in `json_encoder.py` returns `str(o)` for any non-finite float, at any depth, and dumps with `allow_nan=False` so a raw one could never reach the wire.

### Fix

A replacer on the existing `JSON.stringify` call, which reaches every depth, using the same `String(value)` spelling #292 chose. The schema branch moves from `{type: 'null'}` to `{type: 'string'}` so it keeps describing what is actually sent.

Two mutations: dropping the replacer, or reverting the schema branch, each fail the new assertion.

One visible consequence worth calling out. An array like `[NaN, 2]` becomes `["NaN", 2]`, which is heterogeneous, so its schema loses `items` and falls back to a bare `array`. That is the existing heterogeneous-array behaviour rather than anything new here, and the test pins it so the change is not silent.

### DIFF
--- a/.changeset/nested-non-finite.md
+++ b/.changeset/nested-non-finite.md
@@ -1,0 +1,5 @@
+---
+'logfire': patch
+---
+
+Send a non-finite number nested inside an attribute as a string, the way a top-level one already is. `JSON.stringify` writes `NaN` and `Infinity` as `null`, so a nested one arrived indistinguishable from an actual null, and the generated JSON schema described it as `null` to match. Python's encoder returns `str(o)` for a non-finite float at any depth.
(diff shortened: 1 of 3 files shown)


petyosi:
Thanks — fixed on main in 2b0d673 with co-author credit to you. Your replacer and schema change are carried over as-is; the commit combines this with #305 so all number-representability handling lives in one seam (a serializeNumberAttribute helper for the top level plus the single stringify replacer for nested values), since the two PRs restructured the same code independently and conflicted in the test file.

PR #305, opened 2026-09-04
Send an out-of-range integer attribute as a string

ayaangazali:
### Defect

`otlp-transformer`'s `toAnyValue` encodes every integral number as an OTLP `intValue`:

```js
if (!Number.isInteger(value)) return { doubleValue: value }
return { intValue: value }
```

That field is a signed 64-bit integer, so an attribute larger than `2^63 - 1` is sent claiming a width it does not have. `serializeAttributes` passes such a number straight through.

### Evidence

Measured on `3ed526d`:

```
2**64    emitted=18446744073709552000  typeof=number  isInteger=true
-2**64   emitted=-18446744073709552000 typeof=number  isInteger=true
```

Python converts the same values to strings in `prepare_otlp_attribute`, guarded by `OTLP_MAX_INT_SIZE = 2**63 - 1`.

Worth noting because it surprised me: the double nearest the int64 maximum is already out of range. Writing `9223372036854775807` in JS gives `9223372036854775808`, so the boundary sits at `2 ** 63` rather than at the literal maximum.

### Fix

One branch in the number path, comparing through `BigInt` so the boundary is exact rather than approximated in floating point.

The emitted string is `BigInt(value).toString()`, not `String(value)`. For `2**64` those differ: `BigInt` gives `18446744073709551616`, the double's exact value and what Python's `str(int)` produces, while `String` gives the rounded shortest-round-trip form `18446744073709552000`. A mutation swapping them fails.

Three mutations: dropping the range check, dropping the non-finite branch, or swapping `BigInt` for `String` each fail their own assertion.

Deliberately not included: Python also raises a `UserWarning` here. `console.warn` has no per-site deduplication, so on a hot attribute it would repeat once per span. Happy to add it if you would rather match that too.

This touches the same file as #304 but a different function, and the two merge cleanly. I checked rather than assumed.

### DIFF
--- a/.changeset/oversized-int-attribute.md
+++ b/.changeset/oversized-int-attribute.md
@@ -1,0 +1,5 @@
+---
+'logfire': patch
+---
+
+Send an integer attribute outside signed 64-bit range as its exact decimal string. Every integral number is encoded as an OTLP `intValue`, so a larger one claimed a width the field does not have. Python's `prepare_otlp_attribute` converts the same values to strings.
--- a/packages/logfire-api/src/serializeAttributes.test.ts
+++ b/packages/logfire-api/src/serializeAttributes.test.ts
@@ -48,6 +48,25 @@ describe('serializeAttributes', () => {
     expect(JSON.stringify(result)).toBe('{"inf":"Infinity","nan":"NaN","neginf":"-Infinity","ok":1.5,"zero":0}')
   })
 
+  test('sends an integer outside signed 64-bit range as its exact decimal string', () => {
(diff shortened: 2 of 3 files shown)


petyosi:
Thanks — fixed on main in 2b0d673 with co-author credit to you. The bounds check and exact BigInt decimal form are carried over as-is, combined with #304 into one seam for number representability. One addition beyond this PR: Python's prepare_otlp_attribute warns when it converts an oversized integer, so the JS side now warns too (once per process), since the string conversion silently changes the attribute's type otherwise.

PR #308, opened 2026-09-04
Send a BigInt attribute as its exact decimal string

ayaangazali:
### Defect

`JSON.stringify` throws on a BigInt, and `stringifyJsonAttribute` catches that at the top, so any BigInt in an attribute produced `[unserializable]`. A nested one took the whole surrounding object with it, which is the damaging half: one `id` field discards every sibling.

BigInt is the JS analogue of a Python `int` with no width limit, and it is what a user reaches for after hitting the 64-bit ceiling `2b0d673` just addressed.

### Evidence

Measured on `5c3402e`:

```
big       = "[unserializable]"     // 12345678901234567890n
small     = "[unserializable]"     // 7n
nestedBig = "[unserializable]"     // { id: 99n }, the whole object
```

Even `7n` is discarded. Python has no equivalent gap: `isinstance(o, int)` covers arbitrary precision, and `to_json_value` only reaches for a string above the OTLP limit.

### Fix

Send the exact decimal string, at the top level and through the existing replacer for nested values, with the schema branch changed from `undefined` to `{type: 'string'}` so it describes what is sent.

Always a string, rather than a number when small. Python splits because `int` is its native integer type; in JS a BigInt is an explicit opt-in to arbitrary precision, so narrowing the small ones would make the emitted type depend on magnitude, which is harder to consume than one consistent representation.

Three mutations: dropping the top-level branch, the replacer clause, or the schema change each fail.

One existing expectation changes, deliberately. `does not throw when object or array serialization fails` asserted `[unserializable]` for `{value: 1n}`, using a BigInt as one of several ways to trigger a failure. It now expects `{"value":"1"}`. That test's actual guarantee, graceful degradation, is still carried by its `handler` and `symbol` cases, which remain unserializable.

### DIFF
--- a/.changeset/bigint-attribute.md
+++ b/.changeset/bigint-attribute.md
@@ -1,0 +1,5 @@
+---
+'logfire': patch
+---
+
+Send a BigInt attribute as its exact decimal string instead of discarding it. `JSON.stringify` throws on a BigInt, so one anywhere in an attribute produced `[unserializable]`, and a nested one took the whole surrounding object with it.
--- a/packages/logfire-api/src/serializeAttributes.test.ts
+++ b/packages/logfire-api/src/serializeAttributes.test.ts
@@ -332,11 +332,29 @@ describe('serializeAttributes', () => {
       symbol: Symbol('top-level'),
     })
 
-    expect(result['bigintPayload']).toBe('[unserializable]')
+    // A BigInt is representable now, so it no longer stands in for a serialization failure here;
+    // `handler` and `symbol` below still cover the graceful-degradation guarantee this test exists for.
+    expect(result['bigintPayload']).toBe('{"value":"1"}')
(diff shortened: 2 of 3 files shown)


petyosi:
Verified and fixed in 4d26af7, with you credited as co-author — thank you for catching this. Same approach as the last batch: rather than merging, the fix was folded into the file's existing number seam so there is one rule to reason about.

One deliberate divergence from this PR: instead of always sending a BigInt as a string, `serializeBigIntAttribute` follows the rule `serializeNumberAttribute` already documents — the value itself while the double that carries it to OTLP is exact (within ±(2^53−1)), its exact decimal string beyond that. That keeps small BigInts numeric on the wire, matching how Python sends small ints and how our own oversized-number path already behaves; magnitude-dependent typing is the established behavior here, not something this fix introduced. Nested values go through the same function via the existing replacer, and the schema reflects which representation was sent.

PR #310, opened 2026-09-05
Report every failing flush pipeline

ayaangazali:
### Defect

`flushRuntime` awaits the span processors, log record processors and metric readers with `Promise.all`, which rejects on the first failure. A second failing pipeline is never reported, and its rejection is left unhandled.

This is the same shape as `shutdownRuntime`, which was changed to collect each teardown as it settles. `flushRuntime` was not, even though `shutdown()` calls it too, so both paths report at most one flush failure.

### Evidence

Measured on `5c3402e`, with two additional span processors whose `forceFlush` rejects:

```
forceFlush() -> rejected with: first processor flush failed
```

The second processor's error is gone, and under Vitest it surfaces separately as an unhandled rejection.

pydantic/logfire made the equivalent change in #2366: `force_flush` returned only the tracer provider's status and now combines the logger, forwarding and tracer results.

### Fix

The same treatment `shutdownRuntime` already uses. Each pipeline's failure is recorded as it settles, the deadline still bounds the wait, one failure is rethrown as-is, and several are raised together as an `AggregateError`.

Rethrowing a lone failure unchanged matters for the existing contract: `shutdown still closes the SDK when the pre-shutdown flush fails` asserts identity with `rejects.toBe(flushError)`, and the timeout test asserts the `forceFlush timed out` message. Both still hold, and the second new test pins the single-failure case so a future change cannot quietly wrap it.

Two mutations: restoring `Promise.all`, or always wrapping in an `AggregateError` rather than rethrowing a lone failure, each fail.

### DIFF
--- a/.changeset/flush-report-all.md
+++ b/.changeset/flush-report-all.md
@@ -1,0 +1,5 @@
+---
+'@pydantic/logfire-node': patch
+---
+
+Report every failing flush pipeline from `forceFlush()` instead of only the first. The span processors, log record processors and metric readers were awaited with `Promise.all`, which rejects on the first failure, so a second failing pipeline went unreported and its rejection was left unhandled. Multiple failures are now raised together as an `AggregateError`, matching how `shutdown()` already reports its teardowns.
--- a/.changeset/flush-report-all.md
+++ b/.changeset/flush-report-all.md
@@ -2,4 +2,4 @@
 '@pydantic/logfire-node': patch
 ---
 
-Report every failing flush pipeline from `forceFlush()` instead of only the first. The span processors, log record processors and metric readers were awaited with `Promise.all`, which rejects on the first failure, so a second failing pipeline went unreported and its rejection was left unhandled. Multiple failures are now raised together as an `AggregateError`, matching how `shutdown()` already reports its teardowns.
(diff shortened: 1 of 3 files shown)


ayaangazali:
You are right, and the claim was mine to check. Corrected in `d75854b`, in both the comment and the changeset.

`Promise.all` attaches its resolve and reject element functions to every input promise during the initial iteration, so a later rejection is observed and never becomes an unhandled rejection. The caller simply never sees it. The defect and the fix are unchanged; only my explanation of the mechanism was wrong.

Worth saying where the wrong claim came from, since it was not invented: while developing this I did see the second processor's error reported separately by Vitest. That was my own probe leaving an active runtime whose teardown re-flushed the same failing processors, which I fixed with a `shutdown().catch()` in the committed tests. I read that symptom back into `Promise.all` instead of checking the spec.

The wording now says it rejects with the first failure and never surfaces the rest, which is what actually justifies the change.

petyosi:
Verified and fixed in f9dedfa, with you credited as co-author. The defect was real and your evidence solid — one correction to the description: `Promise.all` does attach handlers to every promise, so the second rejection was handled-but-unreported rather than unhandled.

The implementation diverges in shape: rather than duplicating the settle-each-operation-under-a-deadline block that `shutdownRuntime` carries, that pattern is now a single `settleWithDeadline` helper both lifecycle paths use. Same behavior you proposed — a lone failure rethrown unchanged, several raised as an `AggregateError` — with one copy of the mechanism instead of two. Both of your test cases landed alongside it.

PR #311, opened 2026-09-05
Keep a prototype-named Web Vitals metric attribute

ayaangazali:
### Defect

`setPrimitiveAttribute` in `browserMetrics.ts` writes with a plain record assignment. The keys reaching it come from the `attributes` and `defaultAttributes` callbacks, so a `__proto__` key runs the inherited setter and the entry is dropped from the recorded metric.

This is the class `92f0868` and `a68833f` swept. Neither reached `logfire-browser`.

### Evidence

A recorder configured with `attributes: () => Object.fromEntries([['__proto__', 'proto-value'], ['app.route', '/products/:id']])` records the route and loses the other entry. Reverting the fix fails the new assertion.

### Fix

`Object.defineProperty` at that one write, the same convention `ownRecord.ts` uses. I did not reach for the shared helper because it is internal to `logfire-api` and exporting it would be an API addition rather than a fix.

### Scope, and the four sites I did not change

I started from a grep of the same shape across the browser packages and narrowed it by checking reachability, which removed most of them:

- `webVitals.ts` has two identical writes, but every caller passes a hardcoded key such as `'web_vital.value'`, so no user-supplied name reaches them.
- `browserSession.ts` filters keys through `/^[a-z][a-z0-9_]{0,63}$/` before writing, which rejects `__proto__` outright. Names like `constructor` pass that filter but assign as ordinary own properties, since only `__proto__` has an inherited setter.
- `sessionAttributes.ts` in `logfire-session-replay` has the same shape and is left alone while #309 is open in that package.

### DIFF
--- a/.changeset/browser-metric-own-keys.md
+++ b/.changeset/browser-metric-own-keys.md
@@ -1,0 +1,5 @@
+---
+'@pydantic/logfire-browser': patch
+---
+
+Keep a Web Vitals metric attribute whose name is an `Object.prototype` member. Attribute keys come from the `attributes` and `defaultAttributes` callbacks, and a plain record write to a `__proto__` key ran the inherited setter, so that entry was silently dropped from the recorded metric.
--- a/packages/logfire-browser/src/browserMetrics.test.ts
+++ b/packages/logfire-browser/src/browserMetrics.test.ts
@@ -251,6 +251,26 @@ describe('browser metrics runtime', () => {
     expect(getHistogram(expectedInstruments.TTFB.name).records[0]?.value).toBe(300)
   })
 
+  it('keeps a metric attribute named like an Object.prototype member', async () => {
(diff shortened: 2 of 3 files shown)


petyosi:
Verified and fixed in 5a52742, with you credited as co-author. Your reachability analysis of the four sites you set aside was accurate — but the previous sweeps (92f0868, a68833f) took the position that per-site reachability arguments rot: the `webVitals.ts` writes are one refactor away from taking user keys, and the regex guards in `browserSession.ts` / `sessionAttributes.ts` are adjacent code the write shouldn't depend on. So all six same-shape sites were converted, not just the reachable one: `logfire-browser` got its own `setOwn` helper (the core package deliberately doesn't export its `ownRecord` internals, as you noted), and session-replay's single guarded site got an inline own-property definition. Your Web Vitals recorder test landed as the reachable-path regression test.

PR #312, opened 2026-09-05
Give every concurrent eval case its own task run

ayaangazali:
`Dataset.evaluate` runs its cases concurrently (default concurrency is the whole case list), and each case calls `runWithTaskRun` to install its own task-run context. On the first call of a process, that context comes from an `AsyncLocalStorage` instance the module still has to import.

`ensureALS` marked the probe as done with a boolean *before* awaiting the import:

```ts
alsProbeComplete = true
if (!hasAsyncLocalStorage()) return
const mod = await import('node:async_hooks')   // <- everyone else runs past here
alsImpl = new mod.AsyncLocalStorage<TaskRunState>()
```

The first case suspends on the import. Every case that starts while it is in flight sees `alsProbeComplete === true`, returns immediately, still finds `alsImpl === null`, and takes the no-ALS fallback path — the single-slot `fallbackStore` the file itself documents as unsafe to share. Once the import lands and `alsImpl` becomes non-null, `getCurrentTaskRun` switches to reading the ALS store, which those cases were never registered in, so it returns `undefined` for them.

The user-visible effect is silent: `setEvalAttribute` and `incrementEvalMetric` are no-ops when there is no current task run, so attributes and metrics recorded by those cases are dropped from the report. It only affects the first `evaluate` call, and which cases lose their context depends on timing.

Four concurrent runs on `main`, on Node:

```
expected [ 'a', 'b', 'c', 'd' ]
received [ 'a', undefined, undefined, undefined ]
```

The fix caches the probe as a promise instead of a flag, so every caller awaits the same import and sees `alsImpl` set before deciding which path to take. The regression test uses `vi.resetModules()` because the probe only runs once per module instance and the rest of `offline.test.ts` has already warmed it; without the fix it fails as above.

### DIFF
--- a/.changeset/evals-als-probe-race.md
+++ b/.changeset/evals-als-probe-race.md
@@ -1,0 +1,11 @@
+---
+'logfire': patch
+---
+
+Fix concurrent eval cases sharing one task-run context on the first `Dataset.evaluate` call.
+
+The `AsyncLocalStorage` probe recorded "probe started" in a boolean before awaiting its dynamic
+`import('node:async_hooks')`. Every case that began while that import was still in flight saw the
+flag already set, skipped the await, and fell through to the no-ALS fallback slot, so
+`setEvalAttribute` and `incrementEvalMetric` silently recorded nothing for those cases. The probe
+is now cached as a promise that all callers await.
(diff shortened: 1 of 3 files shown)


petyosi:
Verified and fixed in 5269001, with you credited as co-author — this was a real race, well diagnosed, and your promise-cached probe was the idiomatic fix. It landed essentially as you proposed, including the `vi.resetModules()` regression test measuring four concurrent cases on a fresh module instance (confirmed failing with `['a', undefined, undefined, undefined]` before the fix).

PR #315, opened 2026-09-06
Shut the batch span processor down when the console one fails

ayaangazali:
`LogfireSpanProcessor` is the span processor the Node SDK installs by default (`sdk.ts:391`). It wraps the `BatchSpanProcessor` that holds spans on their way to OTLP, and optionally a console processor. Both lifecycle methods awaited them in sequence:

```ts
async forceFlush(): Promise<void> {
  await this.console?.forceFlush()
  return this.wrapped.forceFlush()
}

async shutdown(): Promise<void> {
  await this.console?.shutdown()
  return this.wrapped.shutdown()
}
```

If the console call rejects, the second line never runs. `this.wrapped` is the batch processor, and its `shutdown` is what exports whatever is still queued, so the failure that skips it is exactly the moment those spans are lost.

Against `main`, with the console exporter's `shutdown` rejecting:

```
BatchSpanProcessor.prototype.shutdown calls: 0
```

This repo has settled the same shape twice. `TailSamplingProcessor.runBoth` says it directly:

> Awaiting them in sequence would skip the second one whenever the first rejects, which is exactly when the deferred processor still needs the call.

And f9dedfa gave `flushRuntime` and `shutdownRuntime` a shared settle-and-report seam for the same reason. This processor was not covered by either.

The fix starts both calls, waits for both, rethrows a lone failure unchanged so the existing rejection contract is unaffected, and raises two as an `AggregateError`. Waiting concurrently also stops a slow console flush from eating into the deadline `flushRuntime` now applies.

Being straight about severity: I did not find an input that makes the console side reject on its own today. `LogfireConsoleSpanExporter.shutdown` resolves, and the global `console` swallows write errors, so this is currently a contract gap rather than a live data loss. It matters because the console exporter is the replaceable, I/O-doing half of the pair and the batch processor is the half whose loss is unrecoverable, and because the two precedents above already decided this ordering is wrong.

### DIFF
--- a/.changeset/node-processor-lifecycle.md
+++ b/.changeset/node-processor-lifecycle.md
@@ -1,0 +1,10 @@
+---
+'@pydantic/logfire-node': patch
+---
+
+Shut the batch span processor down even when the console processor fails.
+
+`LogfireSpanProcessor.forceFlush` and `.shutdown` awaited the console processor before the wrapped
+one, so a console failure skipped the batch processor entirely and every span still queued for
+export was dropped. Both calls now settle together, a lone failure is rethrown unchanged, and two
+are raised as an `AggregateError`.
--- a/packages/logfire-node/src/__test__/traceExporter.test.ts
+++ b/packages/logfire-node/src/__test__/traceExporter.test.ts
@@ -1,0 +1,38 @@
+import { BatchSpanProcessor } from '@opentelemetry/sdk-trace-base'
+import { afterEach, describe, expect, it, vi } from 'vite-plus/test'
+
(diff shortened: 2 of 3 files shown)


petyosi:
Verified and fixed in 6f9d7b1, with you credited as co-author — and the severity honesty was appreciated: it is indeed a contract gap today, not live data loss, which is exactly why it was worth closing while cheap. Landed as proposed: both calls start together, a lone failure is rethrown unchanged, two are raised as an `AggregateError`. One structural divergence: the console processor is now injected by the `logfireSpanProcessor` factory instead of being built inside the class constructor, so the lifecycle is testable with plain doubles rather than module mocks.

PR #325, opened 2026-09-09
Add `@pydantic/logfire-agent-control-ai-sdk`, the Vercel AI SDK adapter for Agent Control

DouweM:
Stacked on #321, which adds the framework-neutral core as `@pydantic/logfire-node/agent-control`. This targets `agent-control-core` and will retarget to `main` when that merges.

## What this is

Agent Control lets someone change what a running agent does — its instructions, its model, its model settings, the names and descriptions its tools are advertised under — from the Logfire UI, without a deploy. This is the [Vercel AI SDK](https://ai-sdk.dev) adapter for it.

```bash
npm install @pydantic/logfire-agent-control-ai-sdk
```

```ts
import { anthropic } from '@ai-sdk/anthropic'
import { agentControl } from '@pydantic/logfire-agent-control-ai-sdk'
import { ToolLoopAgent } from 'ai'

const agent = new ToolLoopAgent(
  agentControl({
    settings: {
      id: 'checkout_assistant',
      model: anthropic('claude-fable-5-1'),
      instructions: 'You are a concise checkout assistant.',
      tools: { get_weather },
    },
  })
)
```

That is the whole installation. The config lives in a Logfire variable named `agent__checkout_assistant`, and the first model request publishes a description of the agent as written so the editor shows what it is you are changing. Until someone publishes a value — and any time Logfire is unreachable, or a published value cannot be understood — the agent runs exactly as the code defines it.

It installs as one `LanguageModelMiddleware`, which is the AI SDK's only per-model-request seam and turns out to be enough for all four sections: what reaches it is the full `LanguageModelV4CallOptions` — the system messages, the tools with their resolved JSON Schemas, and every generation setting — and what comes back is where a renamed tool call has to be mapped to the tool the code implements, before the SDK looks the name up and finds nothing. `agentControl({ settings })` wraps a `ToolLoopAgent`'s settings; `agentControl({ model })` wraps a bare model, for `generateText`, `streamText`, or anywhere a `LanguageModel` goes; `agentControlMiddleware({ name })` is the same thing as a plain middleware for a `wrapLanguageModel` call you already have. All three leave the caller's objects untouched.

## Placement: a package, not another subpath

#321's reasoning was that **every `packages/*` entry is a runtime and every feature area is a subpath**, which is why the core is `@pydantic/logfire-node/agent-control`. An adapter is neither, and the deciding constraint is a third-party one: this code imports `ai` and `@ai-sdk/provider`. A subpath of `@pydantic/logfire-node` would put `ai` in the Node SDK's peer dependencies and its typecheck — for a module most of that package's users will never import — and three adapters would put three frameworks there. #321's own docs page already says the three adapters would each be a package for that reason; this is that.

The repository is not actually "packages are runtimes" without exception: `@pydantic/logfire-session-replay` is a package precisely because it holds a third-party dependency (`@rrweb/record`) that `@pydantic/logfire-browser` must not drag in, and it is wired as an **optional peer** of the package that uses it. That is the arrangement to compare against, and the difference is which way the dependency points. Session replay is a leaf the browser package reaches for; an adapter is a layer on top of the core, so the peer runs the other way:

```jsonc
"peerDependencies": {
  "@pydantic/logfire-node": "workspace:^",
  "ai": "catalog:",
  "@ai-sdk/provider": "catalog:",
  "@ai-sdk/provider-utils": "catalog:",
  "@ai-sdk/gateway": "catalog:"   // optional; only for a published model string with no `providers`
}
```

`@pydantic/logfire-node` is a **peer rather than a dependency**, and that is load-bearing rather than tidy. The SDK keeps its variable provider in a module-level singleton, and so do the core's warn-once and publish-once guards; a nested second copy of `@pydantic/logfire-node` would mean the adapter reading a provider the application's `logfire.configure()` never touched, and `resetAgentControl` in a test suite resetting guards nobody holds. A peer makes that a resolution error instead of a silent one. The range is `workspace:^` rather than a pinned `^0.18.24` because the `/agent-control` subpath does not exist until #321 releases; `workspace:^` stamps whatever version is current when the two are published together.

The costs of a package, stated plainly: a new npm publish target and a trusted-publisher setup. Per the release workflow's own comment, a brand-new package cannot use OIDC for its first publish ([npm/cli#8544](https://github.com/npm/cli/issues/8544)) — it needs one manual publish, then its trusted publisher configured on npmjs.com, and CI takes over.

The Mastra and Codex SDK adapters are being opened in parallel and justify their own placement.

## What becomes editable

| Source / canonical field | Support | Notes |
| --- | --- | --- |
| A declared block of `instructions` | **Yes** | Addressed as `system:0`, `system:1`, … or by `providerOptions: { logfire: { id: 'refunds' } }` on the message. Replaced by publishing text, removed by publishing none; the block keeps its position and its own `providerOptions`, an Anthropic cache breakpoint included. |
| Adding a block | **Yes** | Lands at the end of the static group, ahead of the first block the agent recomputes, so it never moves a provider's prompt-cache boundary. |
| A block a `prepareCall` / `prepareStep` hook computes | **No** | Recognized by its text differing from what the code declares, and reported. Replacing it would pin one run's rendering forever. |
| A tool's advertised name, description, and parameter descriptions | **Yes** | Top-level parameters only. Names, types, and requiredness stay code-defined. |
| Narrowing an override to one toolset | **No** | The AI SDK's `tools` is one flat record with no grouping — `client.tools()` merges MCP tools into it too — so there is nothing to match on. |
| A provider-defined tool | **No** | Passes through untouched, and its name is reserved so nothing can be renamed onto it. |
| `model` | **Conditional** | Resolved through `resolveModel`, else `providers`, else the AI Gateway. |
| The eleven canonical settings | **Conditional** | This adapter lowers them onto the request; what the provider then does with one is a separate question, and the live tests below are what settles it. |

Anything published that *this adapter* cannot apply goes through `onUnmatched` — `'warn'` (default, once per process per message), `'ignore'`, or `'error'`. The full table with every condition is in the [package README](https://github.com/pydantic/logfire-js/blob/agent-control-ai-sdk/packages/logfire-agent-control-ai-sdk/README.md).

## When config applies, and precedence

**Per model request** — every step of the tool loop, not once per conversation, so a value saved in Logfire reaches the next model request with no restart and a rollout can move between two steps of one run. The whole request runs inside the resolution's telemetry context, so its spans carry the label that drove it.

Precedence is **code < published < what a run passed explicitly**, and the AI SDK cannot say which is which: by the time a request reaches a language model middleware, the agent's settings and the call's are one object, and the one per-run hook that takes settings (`prepareCall`) is a whole-object transform whose idiom is to spread through everything it did not change. So a run's own values are recognized by comparison against what the code declares, which the two `agentControl` forms supply. That is exact except for a run that sets a key to the value the code already had, where the published value wins. Without `codeSettings` there is no comparison to make and a published value wins every key it names. The README says so out loud.

## What user code sees on a rename

Its own name, everywhere. A rename is a costume the tool wears in front of the model: the request advertises `lookup_weather`, the model calls `lookup_weather`, and the `get_weather` implementation runs with the arguments it sent. `result.steps[].toolResults`, `onToolExecutionStart`, and `response.messages` all say `get_weather`. Three things follow the rename outwards so the model sees one consistent tool set — the definition, a forced `toolChoice` that names the tool, and the `tool-call`/`tool-result` entries the SDK writes into the next step's prompt, which would otherwise alternate names between steps and bust the prompt cache.

## The provider vocabulary is Pydantic AI v2's

The contract's model strings are `provider:model` in Pydantic AI's vocabulary, and the table that translated them was keyed on `google-gla` and `google-vertex` — **ids that no longer exist**. Verified against the installed release: `infer_model('google-gla:gemini-2.5-flash')` raises `Unknown model … Did you mean 'google:gemini-2.5-flash'?`, and `google-vertex:` suggests `google-cloud:`. Every other id the README names (`anthropic`, `openai`, `xai`, `groq`, `mistral`, `deepseek`, `cohere`) resolves.

So:

- `google` (Gemini API) is now the canonical name, and it needs no translation at all — the AI SDK spells it the same way.
- `google-cloud` (Vertex AI) is the canonical name for Vertex, translated to `vertex`, which is the name `@ai-sdk/google-vertex` exports its provider under.
- `google-gla` and `google-vertex` stay as **accepted input aliases**, because a config published against a v1-era agent is still sitting in a Logfire project. Nothing ever writes them, and the table says which are current and which are legacy.

Two things fell out of that:

- **The old `google-vertex` row reported rather than translating**, on the grounds that Vertex had "no conventional AI SDK name". That premise was wrong — `import { vertex } from '@ai-sdk/google-vertex'` is as conventional as `{ anthropic }` — so it now translates. `UNTRANSLATABLE_PROVIDERS` is gone, and with it the branch that refused: a name with no translation is forwarded as it stands, and if nothing serves it the registry or gateway raises, which is caught and reported through `onUnmatched` exactly like every other unresolvable model. A provider package released tomorrow works here on the day it lands.
- **The reverse direction had a real bug.** The baseline named a model by the first segment of its provider id, and Google is one AI SDK namespace and two contract providers: `@ai-sdk/google` reports `google.generative-ai` and `@ai-sdk/google-vertex` reports `google.vertex.<api>`. A Vertex model was therefore published as `google:…`, offering the editor an override that would reach a different service. It now matches on the longest dotted prefix.

## Live provider tests

Every "the provider does X" claim in this adapter's README was read off provider source rather than observed. `src/__test__/live.test.ts` fixes that: eight tests recorded once against real models and replayed offline, with no credentials and no network.

**The recorder.** This repository has no HTTP recording tooling, and both candidates for adding some intercept globally — `nock` patches Node's http stack, `msw` installs a service-worker equivalent. Every `@ai-sdk/*` provider factory already takes a `fetch` of its own, so the recorder is that function: ~130 lines in `src/__test__/cassette.ts`, no dependency, nothing global, no vitest configuration. That matches the standard #321 landed under — the repository as it stands, rather than tooling introduced for one package. Cassettes store no request headers at all (allow-listed to `content-type`) and normalize credential-bearing query parameters out of the URL at both record and replay time, so no key can reach one. Coverage tooling stays off, as it is repository-wide during the Vite+ migration.

Re-record with `node scripts/record-live-cassettes.mjs --env-file <path>`; the script loads the env file into the test subprocess and never prints it.

Models are the cheapest current one per provider, each checked against its provider package's own model-id union rather than recalled: `gpt-5.4-nano`, `claude-haiku-4-5`, `gemini-3.5-flash-lite`.

**What they proved.** A published instruction block replaces the code's on the wire and changes what the model does (it answers `BANANA`). A renamed tool is advertised to Claude under its managed name with its reworded description *and* its reworded parameter description, is called under that name, routes to the `get_weather` implementation with the right arguments, comes back to application code as `get_weather`, and the next turn's history carries `lookup_weather` — through `wrapGenerate` and through `wrapStream` alike. A published `google:` model resolves through `@ai-sdk/google` and answers the request the code-defined OpenAI model would have.

**What they disproved.** `thinking` → `reasoning` is applied and accepted by all three v4 providers — OpenAI as `reasoning.effort`, Anthropic as a `thinking` budget, Gemini as `thinkingConfig.thinkingLevel` — which was the claim the README was least sure of, and it holds. The rest of the settings table did not:

| Setting | OpenAI (Responses) | Anthropic | Gemini |
| --- | --- | --- | --- |
| `max_tokens` | applied | applied, but **raised** to fit the thinking budget (256 → 6656) | applied |
| `temperature`, `top_p` | dropped (reasoning model) | dropped once `thinking` is on | applied |
| `top_k` | dropped | dropped once `thinking` is on | applied |
| `seed` | dropped | dropped | applied |
| `presence_penalty`, `frequency_penalty` | dropped | dropped | **400 `Penalty is not enabled for this model`** |
| `stop_sequences` | dropped | applied | applied |
| `thinking` | applied | applied | applied |
| `parallel_tool_calls` | applied | applied (inverted) | reported unapplied, as designed |

Three consequences, all now in the README:

1. Eight rows that said **Yes** now say **Conditional**, with what each provider was observed to do.
2. "Nothing is dropped in silence" was false. A setting the provider drops is reported by the *AI SDK*, on `result.warnings` as `{ type: 'unsupported', feature: 'topK' }` — not by `onUnmatched`, which never sees it because from this adapter's side the setting was applied. There is a new README section saying which list to read.
3. A published setting can **fail a request outright**, which is the one case that contradicts "a published value can only ever be ignored". `google-penalties.json` is that request, recorded.

**Follow-up I did not take, deliberately.** The middleware could compare `result.warnings` against the keys the published layer actually won and route provider-dropped ones through `onUnmatched`, which would make the "nothing in silence" promise true rather than merely documented. It is a contained change and it closes a real gap, but it is a behavior change beyond porting the package, so it belongs in its own PR with its own review.

## Known limits (all in the README)

A model swap is invisible to AI SDK telemetry, because `wrapLanguageModel` fixes the reported `modelId` at wrap time; the adapter warns once. A model passed as a string cannot be managed and is refused with that message. A per-run `timeout` cannot override a published one — the AI SDK has already merged the run's timeout and its abort signal into one signal — so the two are composed and the shorter wins. Positional block ids move if a hook injects a system message ahead of yours, which the adapter notices and refuses rather than rewriting the wrong block. A tool override invalidates Anthropic's prompt cache once, when you publish. A baseline publish is read-modify-write and can lose a UI edit saved inside that one round trip; `publishBaseline: false` turns it off.

## Review round

Macroscope found five real holes, all now fixed with a regression test each (every one verified by reverting the fix and watching the test fail):

- **Two system messages can carry one instruction id** — a `providerOptions.logfire.id` declared twice, or a hook injecting a message under a name the code already uses. Writing the applied blocks back keyed on id kept only the last slot, so the earlier message was **dropped from the prompt** the moment any config applied, settings-only included. It is a queue per id now; the core addresses both together and in order, which is what makes the queue line up.
- **`codeInstructions: []` is a declaration**, not the absence of one. Reading it as "nothing declared" left the first request with nothing to compare an injected system message against, and made one tenant's rendering addressable.
- **The tool routing tables were ordinary objects**, keyed by names the model and the caller choose. A call to an unadvertised `toString` read back `Object.prototype.toString` and put a function where a tool name goes. Null prototypes.
- **The model resolver cached the settled model, not the in-flight promise**, so two requests arriving on one new id both built a model and split whatever the provider caches between them.
- **An added block could land past the conversation.** A prompt's system messages are not necessarily all at the front — `messages` may carry one and `allowSystemInMessages` lets it through — and the write-back emitted added blocks after the last *slot* rather than after the last block that survived, so removing that trailing block put a managed instruction after the user's own turn. It goes after the last survivor now, and at the front when nothing survived.

One finding I did not take: that calling `model.doGenerate` in place of the handed-in `doGenerate` bypasses the middleware chain. It does not — `wrapLanguageModel` builds that continuation as `() => model.doGenerate(params)` over the very `model` it also passes us, and this middleware declares no `transformParams`, so the two are the same call with different params, inner middleware included. Only a *managed* model is different, which is the point of naming one. The module docstring now says so rather than leaving the next reader to re-derive it.

Macroscope also asked for the cassette recorder to match request bodies and not just method and URL, which was right and immediately earned its keep: it caught that the streaming tool cassette had been recorded from a run that timed out mid-stream, so its second turn had gone out with the code-side tool name and no managed config at all. Re-recorded — the re-rename through `wrapStream` is now proven on the wire rather than only in the request object.

## Review round

Five reported defects reproduced and are fixed, each with a regression test that fails without its fix: a declared instruction id in the reserved `tag:` namespace was dropped from the prompt outright; an id two entries both declared addressed both of them; a tool keyed `__proto__` vanished from a rebuilt record; `raiseSettings` raised `parallel_tool_calls` out of any provider's namespace rather than the serving one; and the README named `LOGFIRE_TOKEN` where a managed variable needs `LOGFIRE_API_KEY`.

Two did not hold up. Instructions do not stop applying across a tool loop — Mastra rebuilds a step's system messages from the agent, not from what the previous step returned, and the two-step test now asserts that nothing is reported on the second step. And a per-run `instructions:` option that *starts* with the agent's own text genuinely cannot be told from that agent plus a block a subsystem added: `ProcessInputArgs` carries no per-run `instructions`, and both land in the same untagged bucket. Documented as a known limit rather than guessed at.

## Tests and gates

102 tests in the new package (94 offline against `MockLanguageModelV4` and the SDK's own `LocalVariableProvider`, 8 replayed from cassettes), and the repository goes from 314 to 416.

Everything CI runs, from the root, green: `vp install --frozen-lockfile`, `vp run --filter "./packages/*" build`, `vp check` (541 files formatted, 338 lint- and type-clean), `vp run --filter "./packages/*" typecheck`, `vp run --filter "./packages/*" test`, and `node scripts/create-npm-token.test.mjs`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01CTCke3iTd1KUvgqwpXxE9f

### DIFF
--- a/packages/logfire-agent-control-ai-sdk/src/middleware.ts
+++ b/packages/logfire-agent-control-ai-sdk/src/middleware.ts
@@ -21,6 +21,14 @@
  * difference between "this agent regressed" and "this agent regressed on the value someone saved at
  * 14:02". It is also what makes the model swap ordinary rather than special: a managed model is just
  * a different object to call with the same params.
+ *
+ * Calling `model.doGenerate` in place of the `doGenerate` we were handed does not step out of the
+ * middleware chain. `wrapLanguageModel` builds that continuation as `() => model.doGenerate(params)`
+ * over the very `model` it also passes us (`ai/src/middleware/wrap-language-model.ts`), and this
+ * middleware declares no `transformParams`, so the two are the same call with different params --
+ * and `model` is already wrapped in whatever middleware sits *inside* this one. Only the managed
+ * model is different, and that is the point of it: a config that names another model is asking for
+ * another model, not for this one with different arguments.
  */
 
 import type { AgentConfig, OnUnmatched } from '@pydantic/logfire-node/agent-control'
@@ -152,9 +160,11 @@ export function agentControlMiddleware(options: AgentControlOptions): LanguageMo
   // model and a `prepareStep` can change the set per step -- which the README's limits say out loud.
   const baselineSource = options.codeInstructions !== undefined && options.codeSettings !== undefined ? 'code' : 'observed'
 
-  // Learned from the first request when the agent did not declare its instructions; see
-  // `CodeInstructions`.
-  let code = declared.length > 0 ? CodeInstructions.declared(declared) : CodeInstructions.unknown()
+  // Learned from the first request only when the agent declared *nothing*; see `CodeInstructions`.
+  // An agent that declares an empty `instructions` has said something -- that every system message
+  // reaching the model came from somewhere else -- and an empty declaration is what says so, where
+  // falling through to `unknown()` would make an injected block addressable on the first request.
+  let code = options.codeInstructions === undefined ? CodeInstructions.unknown() : CodeInstructions.declared(declared)
 
   /** Publish the baseline once, then apply whatever the resolved config changes. */
   async function prepare(params: LanguageModelV4CallOptions, model: LanguageModelV4, config: AgentConfig | null): Promise<PreparedRequest> {
--- a/packages/logfire-agent-control-ai-sdk/src/middleware.ts
+++ b/packages/logfire-agent-control-ai-sdk/src/middleware.ts
@@ -1,0 +1,255 @@
+/**
+ * The hook: one `LanguageModelMiddleware` that applies a managed config to every model request.
+ *
+ * `wrapLanguageModel` is the AI SDK's only per-model-request seam, and it is enough for all four
+ * sections. What reaches it is the full `LanguageModelV4CallOptions` -- the system messages, the
+ * tools with their resolved JSON Schemas, and every generation setting -- which is everything a
+ * managed config patches; what comes back is where a renamed tool call has to be mapped to the tool
+ * the code actually implements, before the SDK looks the name up and finds nothing.
+ *
+ * # Why the rewrite happens in `wrapGenerate` rather than in `transformParams`
+ *
+ * The natural place for a parameter rewrite is `transformParams`, and the natural place to map a
+ * result back is `wrapGenerate` -- but they are two calls, and one model request would then resolve
+ * the variable twice and have to smuggle the routing table from the first to the second.
+ * `wrapGenerate` and `wrapStream`, on the other hand, are already a scope around the whole request:
+ * the `doGenerate` they are handed is literally `() => model.doGenerate(params)`, so calling
+ * `model.doGenerate(ourParams)` in its place transforms and executes in one place.
+ *
+ * That one scope is what `AgentControl.run` needs. Everything inside it carries the resolved label
+ * on its spans, so the trace for a request says which published version drove it -- which is the
+ * difference between "this agent regressed" and "this agent regressed on the value someone saved at
+ * 14:02". It is also what makes the model swap ordinary rather than special: a managed model is just
+ * a different object to call with the same params.
+ */
+
+import type { AgentConfig, OnUnmatched } from '@pydantic/logfire-node/agent-control'
+import { AgentControl, applyInstructions, buildBaseline, warnOnce } from '@pydantic/logfire-node/agent-control'
+import type {
+  LanguageModelV4,
+  LanguageModelV4CallOptions,
+  LanguageModelV4GenerateResult,
+  LanguageModelV4StreamPart,
+  LanguageModelV4StreamResult,
+} from '@ai-sdk/provider'
+import type { Instructions, LanguageModelMiddleware } from 'ai'
+
+import { baselineBlockOf, blockOf, CodeInstructions, instructionMessages, systemSlots, writeSystemMessages } from './instructions'
+import { createModelResolver, modelIdentifier } from './model'
+import type { ModelResolutionOptions, SpecificationVersion } from './model'
+import { canonicalSettings, lowerSettings, providerNamespace, supportedSettings } from './settings'
+import type { CallSettings } from './settings'
+import { applyToTools, renameHistory, renameToolChoice, toolDefs, unrenameContent, unrenameStreamPart } from './tools'
+
+/** How an agent is managed: which variable it reads, and what this adapter is allowed to know. */
+export interface AgentControlOptions extends ModelResolutionOptions {
+  /**
+   * The name of the agent, which picks out the `agent__<name>` Logfire variable holding its config.
+   *
+   * Required and never derived from anything the AI SDK carries. `ToolLoopAgent`'s `id` and
+   * `telemetry.functionId` are both optional and both free-form, so guessing at one would let two
+   * agents share a config, or move an agent onto a different config the day someone renames it.
+   */
+  name?: string
+  /** Read this label instead of letting the variable's rollout choose one. */
+  label?: string
+  /** What to do with a published entry that reaches nothing: `'warn'` (default), `'ignore'`, `'error'`. */
+  onUnmatched?: OnUnmatched
+  /** Set `false` when the Logfire token is read-only or code must not write variable metadata. */
+  publishBaseline?: boolean
+  /**
+   * The instructions the agent declares in code.
+   *
+   * Two things depend on it, and both are about provenance. It tells a declared block from one a
+   * `prepareCall`/`prepareStep` hook injected, which is what decides whether a block is addressable;
+   * and it is the only evidence that a block's text is the agent's own rather than one request's
+   * rendering, which is what decides whether that text may be published into the shared baseline.
+   * Without it every block is published as a seam. `agentControl` passes the agent's own.
+   */
+  codeInstructions?: Instructions
+  /**
+   * The generation settings the agent declares in code.
+   *
+   * Used for precedence -- a request field that differs from this one was set for this run, and a
+   * managed value does not overwrite it -- and for the baseline, so the Logfire editor is shown the
+   * agent's own defaults rather than whatever the first request happened to carry. Without it a
+   * managed setting wins over a per-call one, because by this point the two are indistinguishable.
+   */
+  codeSettings?: CallSettings
+  /**
+   * The specification version of the model this middleware is installed on.
+   *
+   * `wrapLanguageModel` accepts a v2, v3, or v4 model and bridges the older two with a `Proxy` whose
+   * only behavior is to answer `'v4'` to `specificationVersion`, so a middleware cannot ask what it
+   * is really talking to. It matters for exactly one setting: `reasoning` is the call option v4
+   * added, and an older provider is handed it and drops it, so a published `thinking` has to be
+   * reported rather than applied. `agentControl` reads it off the model it is given; a hand-rolled
+   * `wrapLanguageModel` install on a pre-v4 model should pass it. Defaults to `'v4'`.
+   */
+  modelSpecificationVersion?: SpecificationVersion
+}
+
+/** One model request, as this middleware decided to make it. */
+interface PreparedRequest {
+  /** The call options to send, with every managed section applied. */
+  params: LanguageModelV4CallOptions
+  /** The model to send them to: the managed one where a config named one, else the wrapped one. */
+  model: LanguageModelV4
+  /** Advertised tool name to code-side name, for un-renaming what comes back. */
+  routes: Record<string, string>
+}
+
+/**
+ * Require a name rather than inventing one.
+ *
+ * The name is what picks the variable, so an agent that reaches Logfire under a name nobody chose is
+ * worse than one that fails to start.
+ */
+export function requireName(name: string | undefined, hint: string): string {
+  if (name === undefined || name.trim() === '') {
+    throw new Error(
+      'Logfire Agent Control needs an explicit agent name: it is what picks out the `agent__<name>` ' +
+        `variable holding this agent's managed config. ${hint}`
+    )
+  }
+  return name
+}
+
+/**
+ * The middleware that applies an agent's managed config to every request through a model.
+ *
+ * The low-level install, for a `wrapLanguageModel` or `createProviderRegistry` call you already have.
+ * `agentControl` is the same thing with the wrapping done for you, and with the agent's declared
+ * instructions, settings, and model version filled in -- which is what buys the per-run precedence,
+ * the publishable baseline text, and the reporting of a setting the model cannot honour that this
+ * form has to be told about.
+ *
+ * ```ts
+ * const model = wrapLanguageModel({
+ *   model: anthropic('claude-fable-5-1'),
+ *   middleware: agentControlMiddleware({ name: 'checkout_assistant' }),
+ * });
+ * ```
+ */
+export function agentControlMiddleware(options: AgentControlOptions): LanguageModelMiddleware {
+  const name = requireName(options.name, 'Pass `name` to `agentControlMiddleware`, or use `agentControl` on an agent with an `id`.')
+  const control = new AgentControl(name, {
+    ...(options.label === undefined ? {} : { label: options.label }),
+    ...(options.onUnmatched === undefined ? {} : { onUnmatched: options.onUnmatched }),
+    ...(options.publishBaseline === undefined ? {} : { publishBaseline: options.publishBaseline }),
+  })
+  const onUnmatched = control.onUnmatched
+  const resolveModel = createModelResolver(options, (message) => {
+    control.reportUnmatched(message)
+  })
+  const declared = instructionMessages(options.codeInstructions)
+  const wrappedVersion = options.modelSpecificationVersion ?? 'v4'
+  // The baseline describes the agent as written only where the agent said what it is: with its own
+  // instructions and its own settings in hand, every value published is one this adapter can point at
+  // a line of code for. Without them it is a snapshot of whichever request happened to come first,
+  // and the Logfire editor has to be told so rather than shown one run's temperature as a default.
+  // The tool list is the request's either way -- the AI SDK resolves tool schemas on the way to a
+  // model and a `prepareStep` can change the set per step -- which the README's limits say out loud.
+  const baselineSource = options.codeInstructions !== undefined && options.codeSettings !== undefined ? 'code' : 'observed'
+
+  // Learned from the first request when the agent did not declare its instructions; see
+  // `CodeInstructions`.
+  let code = declared.length > 0 ? CodeInstructions.declared(declared) : CodeInstructions.unknown()
+
+  /** Publish the baseline once, then apply whatever the resolved config changes. */
+  async function prepare(params: LanguageModelV4CallOptions, model: LanguageModelV4, config: AgentConfig | null): Promise<PreparedRequest> {
+    const slots = systemSlots(params.prompt, code)
+    code = code.learn(slots)
+
+    const wrappedNamespace = providerNamespace(model.provider)
+    control.publishBaseline(
+      buildBaseline({
+        // A block whose text is not provably the agent's own contributes its seam and not its text;
+        // see `baselineBlockOf`.
+        instructions: slots.map(baselineBlockOf),
+        model: modelIdentifier(model),
+        settings: canonicalSettings(options.codeSettings ?? params, wrappedNamespace, supportedSettings(wrappedNamespace, wrappedVersion)),
+        tools: toolDefs(params.tools),
+      }),
+      { source: baselineSource }
+    )
+
+    if (config === null) {
+      return { params, model, routes: {} }
+    }
+
+    const managed = config.model === undefined ? model : ((await resolveModel(config.model)) ?? model)
+    noteModelSwap(managed, model)
+    // The version of the model this request will actually reach: the one we resolved when a config
+    // named one, and otherwise the wrapped model, whose real version only the caller could tell us.
+    const version = managed === model ? wrappedVersion : managed.specificationVersion
+    const { tools, routes, renames } = applyToTools(params.tools, config, onUnmatched)
+
+    return {
+      model: managed,
+      routes,
+      params: {
+        ...params,
+        prompt: renameHistory(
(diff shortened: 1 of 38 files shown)


packages/logfire-agent-control-ai-sdk/src/middleware.ts
```diff
@@ -0,0 +1,255 @@
...
+        }),
+      },
+    }
+  }
+
+  return {
+    wrapGenerate: async ({ params, model }): Promise<LanguageModelV4GenerateResult> =>
+      control.run(async ({ config }) => {
+        const request = await prepare(params, model, config)
+        const result = await request.model.doGenerate(request.params)
```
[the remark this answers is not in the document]
DouweM: Not a bypass, and I checked rather than assumed: `wrapLanguageModel` builds the continuation it hands us as `() => model.doGenerate(transformedParams)` over the very `model` object it passes alongside it (`ai/dist/index.js`, `doWrap`), and this middleware declares no `transformParams`, so `doGenerate()` and `request.model.doGenerate(params)` are the same call with different params. `model` is already wrapped in whatever middleware sits *inside* this one, so a stacked guardrail or logger still sees the request and the result.

The one case where the inner chain is skipped is a managed model replacing the code-defined one, and that is the semantics of naming another model rather than a bypass of the chain. The module docstring in cd9ad0b now says all of this, so the next reader does not have to re-derive it.

The other five findings were real and are fixed, each with a regression test verified by reverting the fix.

packages/logfire-agent-control-ai-sdk/src/middleware.ts
```diff
@@ -0,0 +1,265 @@
...
+        instructions: slots.map(baselineBlockOf),
+        model: modelIdentifier(model),
+        settings: canonicalSettings(options.codeSettings ?? params, wrappedNamespace, supportedSettings(wrappedNamespace, wrappedVersion)),
+        tools: toolDefs(params.tools),
+      }),
+      { source: baselineSource }
+    )
+
+    if (config === null) {
+      return { params, model, routes: {} }
```
petyosi, line 188: [P2] Use a null-prototype routing map on the fallback path

Both generated and streamed tool calls still pass through `routes[name] ?? name` when no config resolves. For valid tool names such as `constructor`, `toString`, or `__proto__`, this ordinary empty object returns an inherited function/object instead of the original string. Executing the remapping helpers reproduces the corruption; the AI SDK then cannot find the tool implementation using its own-property lookup. This occurs before any configuration is published as well as during fallback. Please use a null-prototype map here, as the managed path already does, or skip remapping.

packages/logfire-agent-control-ai-sdk/src/middleware.ts
```diff
@@ -0,0 +1,265 @@
...
+    }
+
+    const managed = config.model === undefined ? model : ((await resolveModel(config.model)) ?? model)
+    noteModelSwap(managed, model)
+    // The version of the model this request will actually reach: the one we resolved when a config
+    // named one, and otherwise the wrapped model, whose real version only the caller could tell us.
+    const version = managed === model ? wrappedVersion : managed.specificationVersion
```
petyosi, line 195: [P2] Retain the original model specification version across registry conversion

The supported `providers` option accepts `ProviderV3`, but AI SDK 7.0.97's registry converts it through `asProviderV4` and `asLanguageModelV4`. The resulting proxy reports `specificationVersion: 'v4'` while forwarding the older model's methods. After a managed swap to such a provider, this check therefore treats it as supporting `thinking` and writes `reasoning`, which the v3 provider does not support. The request proceeds without an unmatched-setting report even under `onUnmatched: 'error'`. Please retain the original model version before registry conversion. The existing legacy-model tests cover the original wrapped model, but not a registry-resolved replacement.

petyosi (commented):
Review findings from checking the adapter behavior and focused reproductions.

DouweM:
## Parked — handover

**Status.** Draft, parked on sequencing, not on anything in the PR. After talking to David Montague we narrowed the release to the Pydantic AI path only — the `pydantic-ai-harness` capability and the parts of the Logfire Python SDK it exercises — rather than carry maintenance for four more framework adapters before the contract is settled. CI is green on `cd9ad0b5`: `build` passes, Macroscope and Veria are clean (CodeRabbit reports "review rate limited", not a failure; `release` skips on a draft).

**This one stays exercised.** Along with the LangChain adapter (pydantic/logfire#2391), this is a designated cross-provider canary: changes to the cores keep being validated against it, so please do not treat it as frozen or let it rot behind `agent-control-core`.

### What it proves

Every "the provider does X" claim in this adapter's README had been read off provider source. `src/__test__/live.test.ts` fixes that with eight tests recorded once against real models and replayed offline with no credentials — against `gpt-5.4-nano`, `claude-haiku-4-5` and `gemini-3.5-flash-lite`, each checked against its provider package's own model-id union rather than recalled. The recorder is itself worth knowing about: this repository has no HTTP recording tooling, and both candidates intercept globally (`nock` patches Node's http stack, `msw` installs a service-worker equivalent), so the recorder is the `fetch` every `@ai-sdk/*` provider factory already accepts — ~130 lines in `src/__test__/cassette.ts`, no dependency, nothing global, no vitest configuration. Cassettes store no request headers at all (allow-listed to `content-type`) and normalise credential-bearing query parameters out of the URL at both record and replay time.

**Proved:** a published instruction block replaces the code's on the wire and changes what the model does; a renamed tool is advertised to Claude under its managed name with its reworded description *and* parameter description, called under that name, routed to the `get_weather` implementation, returned to application code as `get_weather`, with the next turn's history carrying `lookup_weather` — through `wrapGenerate` **and** `wrapStream`; a published `google:` model resolves through `@ai-sdk/google` and answers what the code-defined OpenAI model would have.

**Disproved:** `thinking` → `reasoning` was the claim the README was least sure of, and it holds on all three v4 providers (OpenAI `reasoning.effort`, Anthropic a thinking budget, Gemini `thinkingConfig.thinkingLevel`). The rest of the table did not:

| Setting | OpenAI (Responses) | Anthropic | Gemini |
|---|---|---|---|
| `max_tokens` | applied | applied, but **raised** to fit the thinking budget (256 → 6656) | applied |
| `temperature`, `top_p` | dropped (reasoning model) | dropped once `thinking` is on | applied |
| `top_k` | dropped | dropped once `thinking` is on | applied |
| `seed` | dropped | dropped | applied |
| `presence_penalty`, `frequency_penalty` | dropped | dropped | **400 `Penalty is not enabled for this model`** |
| `stop_sequences` | dropped | applied | applied |
| `parallel_tool_calls` | applied | applied (inverted) | reported unapplied, as designed |

Three consequences, all now in the README. Eight rows that said **Yes** say **Conditional**. **"Nothing is dropped in silence" was false**: a setting the provider drops is reported by the *AI SDK*, on `result.warnings` as `{ type: 'unsupported', feature: 'topK' }` — not by `onUnmatched`, which never sees it, because from this adapter's side the setting *was* applied. And **a published setting can fail a request outright**, which is the one case that contradicts "a published value can only ever be ignored"; `google-penalties.json` is that request, recorded.

One more finding worth carrying forward: Macroscope asked for the cassette recorder to match request **bodies** and not just method and URL, which immediately earned its keep — it caught that the streaming tool cassette had been recorded from a run that timed out mid-stream, so its second turn had gone out with the code-side tool name and no managed config at all. Re-recorded; the re-rename through `wrapStream` is now proven on the wire rather than only in the request object.

### What v2 changes here

Slice **S1** has landed on `agent-control-core` in logfire-js (`1d60408d`, pydantic/logfire-js#321); this branch is six core commits behind it. S1 breaks every adapter's call sites: the three apply helpers no longer take `onUnmatched` and no longer report — they return `issues`; `UnappliedEntry` → `ApplyIssue`; `reportIssues(policy, issues)` applies the policy once per request; `applySettings` returns an `AppliedSettings` so all three sections share one shape; and `AgentSupport` (sections, destinations, settings) gives TypeScript the filtering it never had.

Verified against `cd9ad0b5`: `src/middleware.ts:196` (`applyToTools`), `:204` (`applyInstructions`), `:214` (the `onUnmatched` handed to `lowerSettings`); `src/settings.ts:197` (`applySettings`) and `:211` (`reportUnapplied(unapplied, …)`); `src/tools.ts:82` (`applyToolDefinitions`).

S1 also lands something this PR explicitly asked for and declined to build: `droppedByProvider(setting, detail)` is now the core's channel for a setting the adapter *did* forward and the provider or its own SDK dropped — exactly `result.warnings`, which nothing read. The follow-up left undone here (compare `result.warnings` against the keys the published layer actually won and route provider-dropped ones through the policy) is now a handful of lines, and it is what makes "nothing in silence" true rather than merely documented.

Still to come: `InstructionPart` replacing `Block`, with independent `dynamic`/`editable`/`publish`/`baseline_text`; **named destinations**; a corrected placement rule (relevant here — an added block already had to learn to land after the last *surviving* block rather than the last slot, because `messages` may carry a system message and `allowSystemInMessages` lets it through); and per-model setting constraints cut from the contract entirely, which is the right home for the table above.

### The one thing that will bite you

**Two open, unanswered P2 findings from @petyosi.** Neither has a reply or a fix, and the first is live before anything is ever published:

1. [The fallback routing path uses an ordinary object](https://github.com/pydantic/logfire-js/pull/325#discussion_r3976253129). Both generated and streamed tool calls go through `routes[name] ?? name` when no config resolves, so a valid tool named `constructor`, `toString` or `__proto__` gets an inherited function or object back instead of its own name, and the AI SDK's own-property lookup then cannot find the implementation. The managed path already uses a null prototype; this path does not. **This fires before any config is published, and on every fallback.**
2. [The model specification version is not retained across registry conversion](https://github.com/pydantic/logfire-js/pull/325#discussion_r3976253133). The `providers` option accepts `ProviderV3`, but AI SDK 7.0.97's registry converts it through `asProviderV4`/`asLanguageModelV4`; the resulting proxy reports `specificationVersion: 'v4'` while forwarding the older model's methods. After a managed swap to such a provider the version check treats it as supporting `thinking` and writes `reasoning`, which the v3 provider does not support — and the request proceeds with no unmatched report even under `onUnmatched: 'error'`. The existing legacy-model tests cover the *wrapped* model, not a registry-resolved replacement.

Also already answered, so don't re-litigate it: calling `model.doGenerate` in place of the handed-in `doGenerate` does **not** bypass the middleware chain — `wrapLanguageModel` builds that continuation as `() => model.doGenerate(params)` over the very `model` it also passes, and this middleware declares no `transformParams` ([thread](https://github.com/pydantic/logfire-js/pull/325#discussion_r3974109727); the module docstring says it too).

### Left unverified / known limits

- A **model swap is invisible to AI SDK telemetry**: `wrapLanguageModel` fixes the reported `modelId` at wrap time. The adapter warns once; nothing fixes it.
- Precedence (code < published < the run's own values) is exact **except** for a run that sets a key to the value the code already had, where the published value wins — the AI SDK has merged the agent's settings and the call's into one object by the time a middleware sees them. Without `codeSettings` there is no comparison at all and a published value wins every key it names.
- A per-run `timeout` cannot override a published one: the AI SDK has already merged the run's timeout and abort signal into one signal, so the two are composed and the shorter wins.
- A baseline publish is read-modify-write and can lose a UI edit saved inside that one round trip; `publishBaseline: false` turns it off.

